PR #29, opened 2021-10-16
Update README.md to reflect the new 'Release' action

Ecorous:


raoulvdberge:
The release action doesn't do enough yet, so the steps are still correct.

PR #185, opened 2022-07-01
Remove guava

raoulvdberge:


refinedstorage2-core-api/src/main/java/com/refinedmods/refinedstorage2/api/core/CoreValidations.java
```diff
@@ -0,0 +1,43 @@
...
+    public static <T> T validateNotNull(@Nullable final T value, final String message) {
+        if (value == null) {
+            throw new NullPointerException(message);
+        }
+        return value;
+    }
+
+    public static void validateNonNegative(final long value, final String message) {
+        if (value < 0) {
+            throw new IllegalArgumentException(message);
```
Technici4n, line 22: Should probably print the negative value, it might be quite insightful.

raoulvdberge: That's the responsibility of the caller, by passing an appropriate message

Technici4n: Not having to format every call to a validation function was the reason I added the following function to the transfer API:
```java
	public static void notNegative(long amount) {
		if (amount < 0) {
			throw new IllegalArgumentException("Amount may not be negative, but it is: " + amount);
		}
	}
```

Issue #234, opened 2022-10-13
Use Gradle version catalogs


raoulvdberge:

bulldog98:
Which one would you prefer settings.gradle.kts or gradle/libs.versions.toml?

raoulvdberge:
@bulldog98 Sorry, I forgot about your comment! I'd prefer it to be regular Groovy in `settings.gradle`. We could then get (mostly) rid of `gradle.properties` (except for the `org.gradle.jvmargs` part).

bulldog98:
@raoulvdberge I had a look at the issue, and the syntax for the catalog change between gradle 7.3 and 7.6, should we update gradle first?

raoulvdberge:
@bulldog98 That's fine - first check if it's possible with Forge/Fabric though to update.

bulldog98:
7.6 does not work, but 7.5 does, and it already uses the new syntax

PR #253, opened 2022-12-27
Switch to version catalog for dependencies

bulldog98:
This PR changes the build according to #234


gradle.properties
```diff
@@ -1,36 +1,5 @@
...
-equalsverifierVersion=3.10.1
-assertJVersion=3.23.1
-mockitoVersion=4.7.0
-jsr305Version=3.0.2
-checkstyleVersion=10.3.2
-# Libraries
-apiGuardianVersion=1.1.2
 # MC
 minecraftVersion=1.19.2
 parchmentVersion=2022.08.21
```
raoulvdberge: Can these be removed now somehow?

refinedstorage2-core-api/build.gradle
```diff
@@ -1,10 +1,8 @@
 archivesBaseName = 'refinedstorage2-core-api'
 
 dependencies {
-    testImplementation "org.junit.jupiter:junit-jupiter-api:${project.junitVersion}"
-    testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:${project.junitVersion}"
-    testImplementation "org.junit.jupiter:junit-jupiter-params:${project.junitVersion}"
-    testImplementation "org.assertj:assertj-core:${project.assertJVersion}"
+    testImplementation testLibs.bundles.junit
```
raoulvdberge: I'm not really a fan of using bundles for junit. The engine should only be provided at runtime, not in implementation.

raoulvdberge (changes requested):

gradle.properties
bulldog98: At the moment I was not able to configure the parchment version via version catalog, opened a bug in the fabric loom gradle plugin about that https://github.com/FabricMC/fabric-loom/issues/782

refinedstorage2-core-api/build.gradle
bulldog98: Will fix that today or tomorrow.

gradle.properties
raoulvdberge: In that case, maybe put a TODO comment on there with a link to your issue.

bulldog98: I had a thought, how to get rid of the versions there, you can also only define a version in the catalog and use it in a string interpolation, we should still add that TODO with the link. Not sure if it should be at the position where the version is defined, or at the position where it's used.

raoulvdberge: That's even better. Probably in the position where it's used,  I think..?

raoulvdberge:
Can you also rebase? I did a minor change to the checkstyle config on develop.

refinedstorage2-core-api/build.gradle
bulldog98: Done

gradle.properties
bulldog98: Could you check if that is ok like that?

raoulvdberge:
The build is failing. `refinedstorage2-network-test` needs to have Junit on the implementation scope.

bulldog98:
> The build is failing. `refinedstorage2-network-test` needs to have Junit on the implementation scope.
Fixed that

gradlew
```diff
@@ -32,10 +32,10 @@
 #       Busybox and similar reduced shells will NOT work, because this script
 #       requires all of these POSIX shell features:
 #         * functions;
-#         * expansions �$var�, �${var}�, �${var:-default}�, �${var+SET}�,
-#           �${var#prefix}�, �${var%suffix}�, and �$( cmd )�;
-#         * compound commands having a testable exit status, especially �case�;
-#         * various built-in commands including �command�, �set�, and �ulimit�.
+#         * expansions «$var», «${var}», «${var:-default}», «${var+SET}»,
```
raoulvdberge, line 35: Could you undo changes in this file?

refinedstorage2-platform-forge/build.gradle
```diff
@@ -4,9 +4,10 @@ buildscript {
             url = 'https://maven.minecraftforge.net'
         }
     }
-    dependencies {
-        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '5.1.+', changing: true
```
raoulvdberge, line 8: Why did you remove this?

refinedstorage2-platform-forge/build.gradle
```diff
@@ -20,12 +21,10 @@ repositories {
     }
 }
 
-apply plugin: 'net.minecraftforge.gradle'
```
raoulvdberge, line 23: Why did you remove this?

raoulvdberge (changes requested):
Almost there. Just a small question and request.

refinedstorage2-platform-forge/build.gradle
bulldog98: Ah that was the old way to apply gradle plugins, that is the legacy way to add a plugin see https://docs.gradle.org/current/userguide/plugins.html#sec:old_plugin_application

refinedstorage2-platform-forge/build.gradle
```diff
@@ -4,9 +4,10 @@ buildscript {
             url = 'https://maven.minecraftforge.net'
         }
     }
-    dependencies {
-        classpath group: 'net.minecraftforge.gradle', name: 'ForgeGradle', version: '5.1.+', changing: true
-    }
+}
+
+plugins {
+    alias(forge.plugins.forgeGradle)
```
bulldog98, line 10: This is the modern and recommended way to use a plugin

raoulvdberge: cool, thanks!

gradlew
raoulvdberge: Heh. Looks like github was just rendering it weirdly.
image

Checked the branch out and looks fine locally.

raoulvdberge (approved):

raoulvdberge:
Thanks!

Issue #251, opened 2022-12-26
Remove log4j usage


raoulvdberge:
Since Minecraft 1.18.2 Minecraft ships Slf4j.

Actions needed:

- We can drop our Log4j dependencies in the `*-api` modules, and depend on Slf4j instead (best to pick same version number as Minecraft)
- We need to start using Slf4j in our `*-api` modules.
  - The tests need a Slf4j implementation like Log4j2, so we still get logging there. 
- We need to start using Slf4j in our Minecraft platform modules.

bulldog98:
Is it ok for me to take this?

raoulvdberge:
Yes, go ahead :-)

bulldog98:
I had a look at the tests. They use log4j currently, but do not configure it, so there is no output from logging. Do we want to keep that behavior, or do we want logging in tests? (Then I suggest using slf4j-simple, since it does not require any configuration)

raoulvdberge:
Yes, slf4j-simple is a good idea! The less configuration, the better.

raoulvdberge:
Oh, another idea: it might be a good idea to add log4j to the disallowed imports in our Checkstyle config as Log4j is still on the classpath on mc-side.

PR #259, opened 2022-12-31
Switch Logging to Slf4j

bulldog98:
This changes the logging to slf4j as discussed in #251, it also implement the checkstyle check, as suggested in https://github.com/refinedmods/refinedstorage2/issues/251#issuecomment-1368095194

### DIFF
--- a/refinedstorage2-grid-api/build.gradle
+++ b/refinedstorage2-grid-api/build.gradle
@@ -1,10 +1,11 @@
 archivesBaseName = 'refinedstorage2-grid-api'
 
 dependencies {
-    implementation libs.log4j.api
+    implementation libs.slf4j.api
     testImplementation testLibs.junit.api
     testImplementation testLibs.junit.params
     testRuntimeOnly testLibs.junit.engine
+    testRuntimeOnly libs.slf4j.impl
     testImplementation testLibs.assertJ
     testImplementation testLibs.mockito
     implementation project(':refinedstorage2-resource-api')
--- a/refinedstorage2-network-api/build.gradle
+++ b/refinedstorage2-network-api/build.gradle
@@ -1,8 +1,8 @@
 archivesBaseName = 'refinedstorage2-network-api'
 
 dependencies {
-    implementation libs.log4j.api
-    testImplementation libs.log4j.core
+    implementation libs.slf4j.api
+    testRuntimeOnly libs.slf4j.impl
     testImplementation testLibs.junit.api
     testImplementation testLibs.junit.params
     testRuntimeOnly testLibs.junit.engine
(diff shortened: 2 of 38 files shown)


bulldog98:
I do not think this one needs a changelog, but I'm unable to add the skip-changelog to it

refinedstorage2-grid-api/build.gradle
```diff
@@ -1,7 +1,7 @@
 archivesBaseName = 'refinedstorage2-grid-api'
 
 dependencies {
-    implementation libs.log4j.api
+    implementation libs.slf4j.api
```
raoulvdberge, line 4: If we have a dependency on slf4j here, we probably need a testRuntimeOnly slf4j-simple here as well to get logging.

refinedstorage2-network-api/build.gradle
```diff
@@ -1,8 +1,8 @@
 archivesBaseName = 'refinedstorage2-network-api'
 
 dependencies {
-    implementation libs.log4j.api
-    testImplementation libs.log4j.core
+    implementation libs.slf4j.api
+    testImplementation libs.slf4j.impl
```
raoulvdberge: This should probably be testRuntimeOnly

settings.gradle
```diff
@@ -92,6 +92,10 @@ dependencyResolutionManagement {
             def log4j = version("log4j", "2.17.0")
             library("log4j-api", "org.apache.logging.log4j", "log4j-api").versionRef(log4j)
             library("log4j-core", "org.apache.logging.log4j", "log4j-core").versionRef(log4j)
```
raoulvdberge: Can we remove log4j from here?

settings.gradle
```diff
@@ -92,6 +92,10 @@ dependencyResolutionManagement {
...
+            def slf4j = version("slf4j", "1.8.0-beta4")
+            library("slf4j-api", "org.slf4j", "slf4j-api").versionRef(slf4j)
+            library("slf4j-impl", "org.slf4j", "slf4j-simple").versionRef(slf4j)
+            bundle("slf4j-impl", ["slf4j-api", "log4j-api", "log4j-core", "slf4j-impl"])
```
raoulvdberge: I don't think a bundle really works here as we want runtimeOnly specifically for slf4j-simple?

raoulvdberge: I think log4j can be removed here as well?

raoulvdberge (changes requested):

settings.gradle
bulldog98: sure, I just forgot it.

settings.gradle
bulldog98: yes, that is also unused, as far as I remember

refinedstorage2-grid-api/build.gradle
bulldog98: Added that

raoulvdberge (approved):

raoulvdberge:
Thanks, good stuff!

Issue #172, opened 2022-06-21
Colored cables


raoulvdberge:
Separates networks.

raoulvdberge:
white;white_dye
orange;orange_dye
magenta;magenta_dye
light_blue;light_blue_dye
yellow;yellow_dye
lime;lime_dye
pink;pink_dye
gray;gray_dye **DEFAULT**
light_gray;light_gray_dye
cyan;cyan_dye
purple;purple_dye
blue;blue_dye
brown;brown_dye
green;green_dye
red;red_dye
black;black_dye

bulldog98:
How should the behavior be, should different colored cables connect to each other?

raoulvdberge:
Different cable colors can't connect, for separation purposes.

bulldog98:
> Different cable colors can't connect, for separation purposes.

Can they connect to any color of grid? I assume they can, since grid colors are only for customization

raoulvdberge:
Yeah, they should be able to connect to any network block (besides different colored cables.)

Edit: I changed my mind on this. See #323

bulldog98:
Should we first implement them with tint on the default cable to get the color, or do we have someone that is able to do art for that, since I'm not gonna lie I'm pretty bad at art.

bulldog98:
Also do we want inworld recoloring? About the recipe do we have a recolor 8 at once or only a recolor 1 cable or both?

raoulvdberge:
> Also do we want inworld recoloring?

Yes, like grids

> About the recipe do we have a recolor 8 at once or only a recolor 1 cable or both?

I'd allow recolor for 1 cable at a time, just like the recoloring recipes for grids etc.

> Should we first implement them with tint on the default cable to get the color, or do we have someone that is able to do art for that, since I'm not gonna lie I'm pretty bad at art.

I'll ask Cyanide if he can create some textures for us. Feel free to start on the impl with the default texture though.

bulldog98:
Yeah I'll start implementing.

bulldog98:
Do we want to keep the current color (black/dark_grey) as default color for cables?

PR #267, opened 2023-01-08
Add colored cables (no assets)

bulldog98:
This adds colored cables according to #172


bulldog98:
Did not build since something with gradle download went wrong, also do we want to change the recipes that use a cable to them accepting the tag?

raoulvdberge:
Yes, please update the recipes to use the tag

Issue #328, opened 2023-02-15
Suppress false positive reduntant suppression warnings


bulldog98:
### Describe your enhancement

We have some instances of `@SuppressWarnings("deprecation")` on functions using registries, since forge deprecates them but fabric needs them. Intellij Idea then complains about these being not needed. `@SuppressWarnings({"deprecation", "RedundantSuppression"})` fixes that.

I'll do a PR for that.

PR #329, opened 2023-02-15
refactor: suppress false positive redundant suppression warnings

bulldog98:
refactoring according to #328


raoulvdberge (changes requested):
Can you fix the branch name?

bulldog98:
> Can you fix the branch name?

Sure I'll have to open a new PR though, since github does not allow for changing the branch name of the from for the PR.

Issue #327, opened 2023-02-15
Disk drive has extra slot


tonyaellie:
### Describe the bug

there is an extra slot at the bottom of the disk drive
![image](https://user-images.githubusercontent.com/46281725/219117291-fd57dfb9-815c-4695-a75a-653f09030c21.png)
![image](https://user-images.githubusercontent.com/46281725/219117338-6b54febd-6b72-4694-987c-99158597903b.png)

### How can we reproduce this bug or crash?

1. Launch fabric 1.19.3 with fabric api and re2 installed
2. place a disk drive
3. observe extra slot 

### What Minecraft version is this happening on?

Minecraft 1.19.3

### What Forge or Fabric version is this happening on?

fabric 0.14.14

### What Refined Storage version is this happening on?

re2 2.0.0-milestone.2.7

bulldog98:
Happens in forge too.

PR #330, opened 2023-02-15
fix: disk drives have 8 slots for putting in disks

bulldog98:
correct #327


bulldog98:
There seems to still be strange behavior if you shift click a disk when all 8 slots are full.

raoulvdberge:
I think you forgot to change it in another place

raoulvdberge (changes requested):
You missed the server-side at AbstractDiskDriveBlockEntity.AMOUNT_OF_DISKS.

Perhaps better to use that constant in the client container initializer as well.

This will also need a changelog entry.

PR #335, opened 2023-02-19
Formalize colored logic

### LINKED ISSUE #323: Formalize connection logic
Colored connection logic is now contained in the cable code. We need to expand it so it will also work for the additional colored variants introduced in #320.

New rules:

- Cable-like parts only connect to:
   - Same colored parts/blocks
   - Default colored parts/blocks
   - Parts/blocks without color
- Full blocks with color connect to:
  - Any colored parts/blocks
  - Parts/blocks without color
- Full blocks without color connect to:
  - Any colored parts/blocks
  - Parts/blocks without color

raoulvdberge:
Fixes #323


refinedstorage2-platform-common/src/main/java/com/refinedmods/refinedstorage2/platform/common/block/entity/AbstractInternalNetworkNodeContainerBlockEntity.java
```diff
@@ -114,11 +115,35 @@ public boolean canPerformOutgoingConnection(final Direction direction) {
 
     @Override
     public boolean canAcceptIncomingConnection(final Direction direction, final BlockState other) {
+        if (!colorsAllowConnecting(other)) {
```
bulldog98, line 118: Could we write this as a strategy, so addons can inject their own custom logic in here?

raoulvdberge: Addons can already write their own strategy by implementing `AbstractNetworkNodeContainerBlockEntity` or `PlatformNetworkNodeContainer` on their block entity.

This class, `AbstractInternalNetworkNodeContainerBlockEntity` is not part of API.

PR #336, opened 2023-02-19
feat: add colored cable like items

### LINKED ISSUE #320: Add colored variants for cable-like items
Importer, exporter, external storage, etc.

Needs #319 

Make sure to respect #323

bulldog98:
closes: #320


refinedstorage2-platform-common/src/main/java/com/refinedmods/refinedstorage2/platform/common/content/Items.java
```diff
@@ -161,14 +162,22 @@ public List<Supplier<ControllerBlockItem>> getRegularControllers() {
...
+    public void addCable(final Supplier<NamedBlockItem> supplier) {
         allCables.add(supplier);
     }
 
-    public List<Supplier<CableBlockItem>> getCables() {
+    public List<Supplier<NamedBlockItem>> getCables() {
         return Collections.unmodifiableList(allCables);
     }
 
+    public void addExporters(final Supplier<NamedBlockItem> supplier) {
```
raoulvdberge:
```suggestion
    public void addExporter(final Supplier<NamedBlockItem> supplier) {
```

refinedstorage2-platform-common/src/main/resources/assets/refinedstorage2/models/item/exporter/base.json
```diff
@@ -1,7 +1,6 @@
 {
   "parent": "block/block",
```
raoulvdberge, line 2: There's no need to put `exporter.json` under a "base" folder. After all, this model file isn't used as a "base" anywhere for other model files. It's just included in the multipart definition.

refinedstorage2-platform-common/src/main/java/com/refinedmods/refinedstorage2/platform/common/item/block/CableBlockItem.java
```diff
@@ -3,7 +3,7 @@
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.level.block.Block;
 
-public class CableBlockItem extends AbstractNamedBlockItem {
+public class CableBlockItem extends NamedBlockItem {
```
raoulvdberge, line 6: `CableBlockItem` can be removed I think?

refinedstorage2-platform-fabric/src/main/java/com/refinedmods/refinedstorage2/platform/fabric/ClientModInitializerImpl.java
```diff
@@ -71,7 +71,9 @@ public void onInitializeClient() {
 
     private void setRenderLayers() {
         BlockRenderLayerMap.INSTANCE.putBlock(Blocks.INSTANCE.getImporter(), RenderType.cutout());
-        BlockRenderLayerMap.INSTANCE.putBlock(Blocks.INSTANCE.getExporter(), RenderType.cutout());
+        Blocks.INSTANCE.getExporter().values().forEach(block ->
+            BlockRenderLayerMap.INSTANCE.putBlock(block, RenderType.cutout())
+        );
```
raoulvdberge:
```suggestion
        Blocks.INSTANCE.getExporter().values().forEach(
            block -> BlockRenderLayerMap.INSTANCE.putBlock(block, RenderType.cutout())
        );
```

raoulvdberge (changes requested):

raoulvdberge:
To fix the ordering issues in the datagen I think that `PROPERTY_BY_DIRECTION` needs to be ordered.

refinedstorage2-platform-common/src/main/resources/assets/refinedstorage2/models/item/exporter/base.json
bulldog98: Under what name and folder do you suggest to put it? Also that was to keep it in the same vain as the cable/base.json for items.

bulldog98:
Do we want the window that is opened on exporter click to also say e.g Cyan Exporter instead of Exporter?

bulldog98:
> To fix the ordering issues in the datagen I think that `PROPERTY_BY_DIRECTION` needs to be ordered.

Yes I had a read at `Map.of` and the implementing `ImmutableMap`  uses a random access order, that is only garantieed to be stable inside one jvm lifetime.

refinedstorage2-platform-common/src/main/resources/assets/refinedstorage2/models/item/exporter/base.json
raoulvdberge: Nvm, this is fine!

raoulvdberge (approved):

PR #467, opened 2024-01-02
fix: grid not retrieving resources from the new master network in a network merge

### LINKED ISSUE #366: Grid menu doesn't handle network changes
### Describe the bug

Watcher listeners become out of date + strategies too.

1. Place a grid
2. Place a cable
3. Place other blocks after the cable
4. Put a destructor destroying the cable from step 2
5. Open the grid quickly, before destructor from step 4 destroys cable from step 2
6. Network is split, see what happens.

Minecraft 1.19.3

### What Forge or Fabric version is this happening on?

latest

latest

### Relevant log output

_No response_

raoulvdberge:
The grid would only sync resources that were part of the (removed) slave network after a network merge because the storages from the master network just "remain" and don't get removed and re-added.

In 1d4d3ae6 an initial fix was made to ensure correct behavior for network splits, by making the grid highest-priority and relying on the moved storages re-adding themselves afterwards.

However, for network merges, it would
completely ignore storages that were already part of the new network as those remain and don't get re-added.

If the storages from the new network remain and don't get re-added, we need to notify the grid of these storages somehow (by re-syncing).

This change adds the re-syncing (replay) necessary for network merges. For network splits, it will still re-sync (replay) but won't actually do anything as the network is newly-created and empty at that point.

Fixes #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 -   The Portable Grid now shows an energy bar in the UI.
 -   The energy bar on creative items now shows the infinity symbol instead of the whole amount.
 
+### Fixed
+
+-   Fixed bug where Grid contents weren't synced properly when a network merge occurs.
+
 ## [2.0.0-milestone.3.2] - 2023-11-03
 
 ### Added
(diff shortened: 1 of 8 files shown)


raoulvdberge:
See #366

Issue #108, opened 2021-12-21
Subnetworks


raoulvdberge:

raoulvdberge:
Need colored cables for this.

PR #521, opened 2024-04-05
refactor: improve contract needed for energy network component

raoulvdberge:
The energy component doesn't need full EnergyStorages since it doesn't need following operations:
* Extracting energy with simulation
* Receiving energy

This will be relevant when implementing the relay since they will add an energy storage to the outer network when energy sharing is enabled.


raoulvdberge:
This enables #108

PR #592, opened 2024-07-04
Port to MDG, make runs work

shartte:


refinedstorage-platform-api/build.gradle
```diff
@@ -2,10 +2,10 @@ plugins {
     id 'java-library'
 }
 
-commonProject()
-
 archivesBaseName = 'refinedstorage-platform-api'
 
+commonProject()
```
shartte: Flipping this around was an accident and is not necessary

Issue #805, opened 2025-02-03
Allow precision up to the millibucket for liquid


jonatjano:
### Describe your enhancement

Some recipes using liquids require less than a bucket worth of the liquid, would be nice to be able to specify these recipes in patterns

PR #829, opened 2025-02-17
fix: fluid amount is to the millibucket

lucasser:
see #805 
fix: allow autocrafting less than a bucket of fluid
fix: item amounts in pattern creation, autocraft monitor, and pattern item tooltip are small size
Refactor:  formatWithUnits parameter qty long to double and add formatThousandth method to format 0.16 buckets as 160m (other functions are not affected, since long is implicitly converted to double)
fix: Fluids displays to 3 decimal places
feat: Fluid renderer appends b after amount to signify bucket

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Changed
 
 -   The filter slots for the Exporter, Constructor and Interface now display whether a resource is missing, the destination does not accept it, the resource cannot be autocrafted due to missing resources, or whether the resource is currently autocrafting.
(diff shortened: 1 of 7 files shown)


Nodrance:
"millionth"?
A millionth is 0.000001
a millibucket is 0.001, which is a thousandth

raoulvdberge:
I'm already working on a patch for these bugs. Setting the min amount to 0 is not a proper fix.

raoulvdberge:
Please, before fixing other unreported bugs, first create an issue so we can handle these things in a structured way.

raoulvdberge:
Btw, the reason for this is that I don't like random code being dumped on me. I'd prefer to first discuss if we want it in RS before you put the effort in writing a PR.

That applies to working on existing issues too btw, always a good idea to join the discord and discuss with me first before picking something up as for example I was already working on that bug in this PR.

Issue #830, opened 2025-02-17
Fluids amount displays to the 0.1 and not to 0.001


lucasser:
### Describe the bug

The fluid amount is rounded to the nearest 0.1 bucket, thus displaying very small amounts of fluid as 0.

### How can we reproduce this bug or crash?

1. Insert less than 0.1 buckets of fluid into storage.
2. the tooltip displays as 0.

### What Minecraft version is this happening on?

Minecraft 1.21.1

### What NeoForge or Fabric version is this happening on?

latest

### What Refined Storage version is this happening on?

latest

### Relevant log output

```shell

```

PR #831, opened 2025-02-17
fix: format fluid amount less than 1 in millibuckets

lucasser:
see #830

### DIFF
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/util/IdentifierUtil.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/util/IdentifierUtil.java
@@ -88,33 +88,21 @@ public final class IdentifierUtil {
         return prefix + id.getNamespace() + "." + fixedPath;
     }
 
-    public static String formatWithUnits(final long qty) {
-        if (qty >= 1_000_000_000) {
-            return formatBillion(qty);
-        } else if (qty >= 1_000_000) {
-            return formatMillion(qty);
-        } else if (qty >= 1000) {
-            return formatThousand(qty);
+    public static String formatWithUnits(final double qty) {
+        if (qty < 0.001) {
+            return "0";
         }
-        return String.valueOf(qty);
-    }
 
-    private static String formatBillion(final long qty) {
-        return FORMATTER_WITH_UNITS.format(qty / 1_000_000_000D) + "B";
-    }
-
-    private static String formatMillion(final long qty) {
-        if (qty >= 100_000_000) {
-            return FORMATTER_WITH_UNITS.format(Math.floor(qty / 1_000_000D)) + "M";
-        }
-        return FORMATTER_WITH_UNITS.format(qty / 1_000_000D) + "M";
-    }
-
-    private static String formatThousand(final long qty) {
-        if (qty >= 100_000) {
-            return FORMATTER_WITH_UNITS.format(Math.floor(qty / 1000D)) + "K";
-        }
-        return FORMATTER_WITH_UNITS.format(qty / 1000D) + "K";
+        return switch ((int) Math.floor(Math.log10(qty) / 3)) {
+            case -1 -> FORMATTER_WITH_UNITS.format(qty * 10e2) + "m";
+            case 0 -> FORMATTER_WITH_UNITS.format(qty >= 100 ? Math.floor(qty) : qty);
+            case 1 -> FORMATTER_WITH_UNITS.format(qty >= 10e4 ? Math.floor(qty / 10e2) : qty / 10e2) + "K";
+            case 2 -> FORMATTER_WITH_UNITS.format(qty >= 10e7 ? Math.floor(qty / 10e5) : qty / 10e5) + "M";
+            case 3 -> FORMATTER_WITH_UNITS.format(qty >= 10e9 ? Math.floor(qty / 10e8) : qty / 10e8) + "B";
+            case 4 -> FORMATTER_WITH_UNITS.format(qty >= 10e12 ? Math.floor(qty / 10e11) : qty / 10e11) + "G";
+            case 5 -> FORMATTER_WITH_UNITS.format(qty >= 10e15 ? Math.floor(qty / 10e14) : qty / 10e14) + "T";
+            default -> "∞";
+        };
     }
 
     public static String format(final long qty) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,9 +15,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 -   The filter slots for the Exporter, Constructor and Interface now display whether a resource is missing, the destination does not accept it, the resource cannot be autocrafted due to missing resources, or whether the resource is currently autocrafting.
 
-### Fixed 
+### Fixed
 
 -   Fixed coloured importers, exporters, and external storages not dropping when broken.
+-   Fixed fluids displaying to 0.1 buckets and not 0.001.
 -   Fixed Pattern Grid processing matrix input slots not mentioning "SHIFT click to clear".
 -   Fixed amount buttons not working in the Autocrafting Preview when the amount is "0" after clicking the "Max" button.
 -   Fixed not being able to insert resources in a Grid when clicking on something that is autocraftable.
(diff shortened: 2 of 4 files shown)


raoulvdberge:
Can you undo the formatting changes?

lucasser:
What do you mean?

raoulvdberge:
If you check the diff, there are quite a bit of style changes. (8 spaces instead of 4)

lucasser:
All fixed

raoulvdberge:
The formatter with units still seems to only use 1 digit after comma?

lucasser:
No, its correct, or at least works as I intended. If there is less than a bucket of fluid, we display in mb by appending m to the value (the green fluid)
Screenshot From 2025-02-19 01-04-39
If I hover over any of the items, the value is displayed to 3 sig figs, but if there is more than 1 bucket of the fluid, we keep it at 1 bucket, since otherwise the inventory becomes too crowded.
Screenshot From 2025-02-19 01-11-24

raoulvdberge:
I don't like the 'm' suffix, why can't we display 0.016?

lucasser:
It looks cleaner, especially with the b identifier. 0.001 vs 1mb. I think having it in millibuckets makes more sense.

RoboXGamer:
for sure the 1mb looks better than 0.001

lucasser:
And it also looks better in patterns

raoulvdberge:
I don't think the concept of "millibuckets" exists in Fabric (1/1000th of a bucket = 1mB on NeoForge, but the fraction is different on Fabric).

lucasser:
I went through a bunch of fabric mods, the base fabric fluid measurement is in Units, with 81000 units to a bucket, but most mods (create, ad astra, others) just put millibuckets with 1000mb -> 1b, so I think it's fine.
Screenshot From 2025-02-19 07-33-24
Also, millibuckets make sense, it's 1/1000 of a bucket.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/util/IdentifierUtil.java
```diff
@@ -88,29 +88,42 @@ private static String getTagTranslationKey(final String prefix, final ResourceLo
         return prefix + id.getNamespace() + "." + fixedPath;
     }
 
-    public static String formatWithUnits(final long qty) {
+    public static String formatWithUnits(final double qty) {
```
raoulvdberge, line 91: please use early return here  instead of a mutable variable. If checkstyle complains about the amount of returns, split up in multiple methods.

lucasser: Is there a specific reason?

lucasser: For future reference

lucasser:
Alright, should be all good now.

raoulvdberge:
Please do not do map lookups for doing a simple formatting operation.

lucasser:
So how would I split up the formatters, since I can't have too many returns in one method? Do I do if value > some cutoff -> got to one method, else go to another?

RoboXGamer:
> So how would I split up the formatters, since I can't have too many returns in one method? Do I do if value > some cutoff -> got to one method, else go to another?

cant you simply do a switch statement?

lucasser:
The code checker doesn't like more that 4 returns per method, and bossman said to return in every case, not set a variable and return it at the end.

RoboXGamer:
> The code checker doesn't like more that 4 returns per method, and bossman said to return in every case, not set a variable and return it at the end.

```java
public static String formatWithUnits(final double qty) {
    if (qty < 0.001) return "0"; // Tiny values
    if (qty < 1.0) return FORMATTER_WITH_UNITS.format(qty / 0.001) + "m"; // Milli range
    
    // For qty >= 1, calculate tier (0 = "", 1 = "K", 2 = "M", etc.)
    double tierValue = qty;
    int tier = 0;
    while (tierValue >= 1000.0 && tier < 7) {
        tierValue /= 1000.0;
        tier++;
    }
    if (tier >= 7) return "∞"; // Beyond "a lot"

    // Units: "", "K", "M", "B", "G", "T", "a lot"
    String units = " KM BGTa lot";
    double value = (qty < Math.pow(1000.0, tier + 1) / 10.0) ? Math.floor(tierValue) : tierValue;
    return FORMATTER_WITH_UNITS.format(value) + units.substring(tier, tier + (tier == 6 ? 5 : 1)).trim();
}
```
could this work?

lucasser:
Do you know what the concern with map is?

RoboXGamer:
> Do you know what the concern with map is?

i dont know exactly but the map would be overkill for such a small issue and memory expensive then compared to this simple loop method
the fastest and memory efficient would be to have switch case and return but cant because of 4 being the max

lucasser:
What about having one variable, setting it inside the switch, and returning at the end?

RoboXGamer:
> What about having one variable, setting it inside the switch, and returning at the end?

that could also work but he said in his above message to not use mutable variable and prefers early return option
he also said could use multiple methods but i think that would make it not readable code

lucasser:
But why no mutable variable? Wouldn't that be the most readable.

jonatjano:
> ```java
>     ...
>     double tierValue = qty;
>     int tier = 0;
>     while (tierValue >= 1000.0 && tier < 7) {
>         tierValue /= 1000.0;
>         tier++;
>     }
>     ...
> }
> ```

could simplify that loop with
```
int tier = log10(tierValue) / 3
```

RoboXGamer:
i also think cant we simply use this syntax 
```java
return switch (tier) {
    case 0 -> "";
    case 1 -> "K";
    default -> "M";
};
```

raoulvdberge:
Just split it up in multiple methods like I suggested initially? No need to over engineer this.

raoulvdberge:
And the reason why I said use early returns and no mutable variables is because I prefer pure functions. Easier to read. There's simply no need for mutability here.

lucasser:
But then that's so much redundancy. Instead of running through the cases and coming up with a multiplier and a unit and calling FORMATTER_WITH_UNITS.format(qty * multiplier) + unit; you have to add the call on every line which is redundant. Especially since i want to add G and T as allowed units

jonatjano:
isn't something like this the cleanest you can get ?
```java
public static String getSuffix(long number) {
	int tier = (int)(Math.log10(number)) / 3;
	return switch (tier) {
		case 0 -> "";
		case 1 -> "K";
		case 2 -> "M";
		case 3 -> "G";
		case 4 -> "T";
		default -> "∞";
	};
}
```

raoulvdberge:
This code has to be performant. It was working perfectly fine before. No need to overhaul the entire thing

lucasser:
I got it working either like this with a suffix switch

````java
public static String formatWithUnits(double qty) {
      if (qty < 0.001) {
          return "0";
      }

      if (qty < 1) {
          return FORMATTER_WITH_UNITS.format(qty * 10e2) + "m";
      }

      int tier = (int) Math.floor(Math.log10(qty) / 3);
      qty /= Math.pow(10, tier * 3);
      return FORMATTER_WITH_UNITS.format(qty >= 100D ? Math.floor(qty) : qty) + getSuffix(tier);
}
````

or like this without one
````java
public static String formatWithUnits(double qty) {
      return switch ((int) Math.floor(Math.log10(qty) / 3)) {
            case -1 -> FORMATTER_WITH_UNITS.format(qty * 10e2) + "m";
            case 0 -> FORMATTER_WITH_UNITS.format(qty >= 100 ? Math.floor(qty) : qty);
            case 1 -> FORMATTER_WITH_UNITS.format(qty >= 10e2 ? Math.floor(qty / 10e2) : qty / 10e2) + "K";
            case 2 -> FORMATTER_WITH_UNITS.format(qty >= 10e5 ? Math.floor(qty / 10e5) : qty / 10e5) + "M";
            case 3 -> FORMATTER_WITH_UNITS.format(qty >= 10e8 ? Math.floor(qty / 10e8) : qty / 10e8) + "B";
            case 4 -> FORMATTER_WITH_UNITS.format(qty >= 10e11 ? Math.floor(qty / 10e11) : qty / 10e11) + "G";
            case 5 -> FORMATTER_WITH_UNITS.format(qty >= 10e14 ? Math.floor(qty / 10e14) : qty / 10e14) + "T";
            default -> "∞";
}
````

lucasser:
Is this better?

raoulvdberge:
Yes, this looks much better. Thank you

Issue #832, opened 2025-02-17
Display 'b' after fluid amount to signify bucket


lucasser:
### Describe your enhancement

Append the char 'b' to the fluid amount render to display units.

PR #833, opened 2025-02-17
feat: display 'b' after fluid amount

lucasser:
see #832

### DIFF
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/resource/ItemResourceRendering.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/resource/ItemResourceRendering.java
@@ -38,10 +38,7 @@ public enum ItemResourceRendering implements ResourceRendering {
 
     @Override
     public String formatAmount(final long amount, final boolean withUnits) {
-        if (!withUnits) {
-            return format(amount);
-        }
-        return formatWithUnits(amount);
+        return !withUnits ? format(amount) : formatWithUnits(amount);
     }
 
     @Override
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 
 -   Autocrafting Upgrade
+-   Display 'b' after fluid amount inside grid
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


raoulvdberge:
You're only doing this when we request the amount with units, but have you thought about whether this should work when requesting the amount *without* units? 

If not, please check the usages of the `formatAmount` method and check if it makes sense.

lucasser:
It shouldn't work when you request the amount without units, since I assume that no units are requested for calculations like crafters. Now it should also be more correct, since the character is appended inside the formatAmount method, when units are requested.

raoulvdberge:
No, it doesn't use these methods for calculations. Please check the usages of the formatter without units, it's used in GUIs.

lucasser:
Again, this is working according to what I expected. When we display the value just in the grid, it shows a unit, but when you hover over it, I don't think the unit is necessary. If you disagree, I can change it.

raoulvdberge:
image

Shouldn't we show  'b' on disk tooltips or the storage monitor?

lucasser:
Got it. Ill fix this once I sort out the other PR.

raoulvdberge:
Can you remove these UNIT constants? There is only 1 usage in each class and for items it's even just empty.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/resource/ItemResourceRendering.java
```diff
@@ -38,10 +38,7 @@ private ItemStack getStack(final ItemResource itemResource) {
 
     @Override
     public String formatAmount(final long amount, final boolean withUnits) {
-        if (!withUnits) {
-            return format(amount);
-        }
-        return formatWithUnits(amount);
+        return (!withUnits ? format(amount) : formatWithUnits(amount)) + "";
```
raoulvdberge: remove the parenthesis and `+ ""`

raoulvdberge (changes requested):

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/resource/ItemResourceRendering.java
lucasser: done

raoulvdberge (approved):

PR #849, opened 2025-02-20
fix: correct units display

### LINKED ISSUE #848: Item formatting units are incorrect
### Describe the bug

Just realized that I added the unit B into the unit formatter. I will fix this as soon as I can, it shouldn't be hard.

1. Insert a billion of an item

Minecraft 1.21.1

Latest

Latest

### Relevant log output

```shell

```

lucasser:
Fix #848


raoulvdberge:
Why is k suddenly lowercase?

lucasser:
Because thats how metric is formatted, and i didnt realize
Screenshot From 2025-02-20 13-12-27

lucasser:
That is also how it's formatted in mekanism

Mekanism supports up to 2^64-1 or roughly 1.8x10^19 Joules of energy in any regular container. This would be displayed as 18 EFE in game (exa-FE).

The list of prefixes for energy, sorted by size from the smallest to largest, is :

f < p < n < µ < m < [nothing] < k < M < G < T < P < E < Z < Y

PR #855, opened 2025-02-22
fix: pattern grid dropping items

### LINKED ISSUE #851: Pattern Grid drops Items and Pattern dont work after that
### Describe the bug

After placing a Pattern grid, creating a Pattern and not clearing Items in the Pattern Grid you can break the Pattern Grid and the Items will drop. Another problem is that the created Pattern cant be placed in a Crafter or when it is already placed and you remove it you can't put it back in the Crafter.
If you clear or modify the Items in the Pattern Grid al works as expected.

**Problem 1**

1. Place a Pattern Grid
2. Create a Pattern
3. Dont clear the Grid
4. Break the Grid
5. Now the Items should drop

--- 
**Problem 2**

1. Place a Pattern Grid
2. Create a Pattern
3. Dont clear the Grid
4. Break the Grid
5. Try to put the Pattern in a Crafter

Minecraft 1.21.1

Fabric Loader: 0.16.10

refinedstorage-fabric-2.0.0-milestone.4.13.jar / main (1a5e858)

### Relevant log output

```shell
The problem is not visible in the logs.
```

lucasser:
fix #851 
Changed crafting grid to put back the items into storage when broken

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 -   Fixed item count rendering behind the item in the Interface slots.
 -   Fixed pickaxes not influencing break speed of blocks.
 -   Fixed non-autocraftables Grid view filter still showing autocraftable resources as they get inserted.
+-   Fixed pattern grid dropping items for pattern
+-   Fixed pattern dropped from broken pattern grid
 -   Fixed autocrafting tasks being completed when not all processing outputs have been received yet.
 -   Fixed items not stacking when externally interacting with an Interface.
 -   Fixed Constructor in default scheduling mode not trying other filters when resource is not available.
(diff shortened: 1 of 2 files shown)


raoulvdberge:
No, the crafting grid should drop the matrix on the ground and  not dump items in storage.

lucasser:
Why?

raoulvdberge:
Because it is the logical behavior in minecraft to drop stuff on the ground. Nothing in RS moves stuff in storage when breaking a block.

raoulvdberge:
Did you address the second problem from the bug report?

lucasser:
It got fixed when i changed the first. I made sure

lucasser:
To me at least, the items in the crafting grid are in the network. Just in a different part of it (with crafting capabilities), so if i break the access point to the network, the items should stay in the network since i never "took them out".

raoulvdberge:
> It got fixed when i changed the first. I made sure

What was the cause?

lucasser:
No clue. When i was testing, i could not put the pattern into the crafter, but when i removed the loop to drop the contents, the issue was gone

raoulvdberge:
Can  you change it so it drops the items instead?
I'd also need to know what the cause of the other problem is.

lucasser:
Gonna try to figure it out

HrPlanlos:
Sorry for the question, but when I create the matrix, i.e. enter the recipe that the pattern should have, the items I place are not consumed, right?
If this is the case, why should they be dropped or go into the system, they will be duplicated, right?

lucasser:
The dropping vs going into the system was about the crafting grid

Issue #883, opened 2025-03-04
Sort Tooltip for Upgrades


starforcraft:
### Describe your enhancement

![Image](https://github.com/user-attachments/assets/9dde9f93-4eff-411f-9e8e-32c6ce4149aa)

It would be pretty neat if the upgrades were sorted. For example in this case:
1. Fortune I Upgrade
2. Fortune II Upgrade
3. Fortune III Upgrade
4. Silk Touch Upgrade
5. Speed Upgrade

PR #895, opened 2025-03-18
chore: ordered tooltips for upgrade slots

HrPlanlos:
The order of the tooltips for the upgrade slots now follows the order in which they were specified in the AbstractModInitializer class

Refs: #883


raoulvdberge:
Nice!

HrPlanlos:
I changed the commit message, hope it is correct now

PR #907, opened 2025-03-28
feat: math expressions in input fields

### LINKED ISSUE #199: Allow math expressions in amount screens
See https://github.com/refinedmods/refinedstorage/issues/2728.

This includes detectors!

lucasser:
fix #199 
Also adds new translations for number parser errors.


lucasser:
Expression parser is updated for tree preview and ready for merge.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -40,6 +41,8 @@ public abstract class AbstractAmountScreen<T extends AbstractContainerMenu, N ex
     protected ActionButton cancelButton;
     @Nullable
     protected EditBox amountField;
+    @Nullable
+    protected N amount;
```
raoulvdberge: Please don't keep extra state. Amount screens need to get the last amount from the text field.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -166,25 +165,50 @@ private void addAmountField() {
...
+        this.amount = (data.getValue().isEmpty()) ? null : data.getValue().get();
         amountField.setTextColor(valid ? 0xFFFFFF : 0xFF5555);
+        setToolTip(data.getTooltip(), amountField);
+    }
+
+    protected void setToolTip(final String text, @Nullable final AbstractWidget element) {
+        final Component tooltip;
+        if (element != null) {
+            //tooltips have to be separate to show min and max values
+            if (text.startsWith("resource_amount_input.too_big")) {
```
raoulvdberge: Passing errors as a string like this is a big no-go. Why do we even show tooltips for too big and too small? Seems like edge cases.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractSingleAmountScreen.java
```diff
@@ -21,7 +21,11 @@ protected AbstractSingleAmountScreen(final T containerMenu,
                                          final Component title,
                                          final Double initialAmount,
                                          final Supplier<Double> minAmount) {
-        super(containerMenu, null, playerInventory, title,
```
raoulvdberge: why the formatting changes?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AmountOperations.java
```diff
@@ -4,11 +4,50 @@
...
 
 public interface AmountOperations<N extends Number> {
-    String format(N value);
+    String format(@Nullable N value);
 
-    Optional<N> parse(String value);
+    ReturnValue<N> parse(String value);
 
-    Optional<N> validate(N amount, @Nullable N minAmount, @Nullable N maxAmount);
+    default ReturnValue<N> validate(final String amount,
```
raoulvdberge: Why is this called validate if it parses numbers?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/DoubleAmountOperations.java
```diff
@@ -26,44 +24,28 @@ private DoubleAmountOperations() {
     }
 
     @Override
-    public String format(final Double value) {
-        return DECIMAL_FORMAT.format(value);
+    public String format(@Nullable final Double value) {
+        return (value == null) ? "" : DECIMAL_FORMAT.format(value);
```
raoulvdberge: Why is this nullable?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/IntegerAmountOperations.java
```diff
@@ -1,57 +1,39 @@
...
-                                      @Nullable final Integer minAmount,
-                                      @Nullable final Integer maxAmount) {
-        final boolean minBoundOk = minAmount == null || amount >= minAmount;
-        final boolean maxBoundOk = maxAmount == null || amount <= maxAmount;
-        return minBoundOk && maxBoundOk ? Optional.of(amount) : Optional.empty();
-    }
-
-    @Override
-    public Integer changeAmount(@Nullable final Integer current,
+    public Integer changeAmount(final Integer current,
```
raoulvdberge: Why remove the nullability here?

raoulvdberge (changes requested):

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AmountOperations.java
lucasser: It used to be called validate. Should i change it?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/DoubleAmountOperations.java
lucasser: Value can be null if an empty string is entered. Can be checked for earlier, but i thought to do a single check here instead of at each input field.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: I think it's good user feedback. If you type something, you'd like to know why it doesn't work, especially for setting crafting amounts.
I use this to return error states. The evaluator throws an exception which then gets converted to a ReturnValue, which has an optional value and or a tooltip to show the user, which is either =number or the error state.
I'm doing this, so the screen can just take that tooltip, check if it's an error, and translate it without having to do an exception throw there.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: Why not track the calculated value instead of recalculating every time you want the number. Once you calculate you need to store that result somewhere. Also, you don't need to keep it persistent for world reloads, so it doesn't have to be saved anywhere permanent.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/DoubleAmountOperations.java
raoulvdberge: Well I would not expect to call format if there is no value, in the first place.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AmountOperations.java
raoulvdberge: yep

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: I think it's a bit too granular feedback. Can you remove it for now and see how it feels?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: Because this is state that is bound to become desynced with the input box

lucasser:
Should i account for initial and reset values in the configuration possibly being null?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: https://github.com/user-attachments/assets/06b9f010-6a35-43f2-9f30-565b3f51cd12

This is what happens if you don't feedback. It's confusing. Alternatively it can just be a red text which is still confusing. Say you want to do 128 cobble -> 128 stone or a mek reactor recipe with 1k glass and you enter it and it just fails. Why? you have no idea.

lucasser: i can move this check to a different place, AbstractAmountScreen::checkValue is probably a more appropriate place for it

raoulvdberge: Can you make it just red text, similar to how it's working right now? The tooltip saying `=0` could stay.

lucasser: image
like this?

lucasser: How should i pass errors?

raoulvdberge: Why is it not showing a tooltip with the calculation?

lucasser: image
This better?

raoulvdberge: yes, looks great!

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -140,7 +141,6 @@ private void addCancelButton(final int x, final int y) {
 
     private void addAmountField() {
         final Vector3f pos = configuration.getAmountFieldPosition();
-        final String originalValue = amountField != null ? amountField.getValue() : null;
```
raoulvdberge, line 143: Why is this removed now? What happens if you resize the screen, is the amount still there?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/autocrafting/preview/AutocraftingPreviewScreen.java
```diff
@@ -929,12 +933,12 @@ protected void reset() {
     }
 
     @Override
-    protected boolean confirm(final Double amount) {
+    protected boolean confirm(final Double value) {
```
raoulvdberge, line 936: Why rename this variable?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -166,25 +162,39 @@ private void addAmountField() {
...
             tryConfirm();
         }
         amountField.setTextColor(valid ? 0xFFFFFF : 0xFF5555);
+        setToolTip(data.getTooltip(), amountField);
+    }
+
+    protected void setToolTip(final String text, @Nullable final AbstractWidget element) {
+        final Component tooltip;
+        if (element != null) {
+            if (text.startsWith("resource_amount_input")) {
```
raoulvdberge, line 191: I don't like this string comparison.

refinedstorage-common/src/main/resources/assets/refinedstorage/lang/en_us.json
```diff
@@ -279,6 +279,14 @@
...
   "gui.refinedstorage.exporting_indicator.autocrafting_was_started": "Autocrafting was started",
   "gui.refinedstorage.exporting_indicator.autocrafting_missing_resources": "Could not autocraft, missing resources",
+  "gui.refinedstorage.resource_amount_input.no_value_before_unit": "Enter value before multiplier unit",
+  "gui.refinedstorage.resource_amount_input.invalid_unit": "Unrecognized multiplier unit",
+  "gui.refinedstorage.resource_amount_input.syntax_error": "Syntax error",
+  "gui.refinedstorage.resource_amount_input.unrecognized_char": "Unrecognized character",
+  "gui.refinedstorage.resource_amount_input.too_big": "Too big\nMaximum is %s",
+  "gui.refinedstorage.resource_amount_input.too_small": "Too small\nMinimum is %s",
+  "gui.refinedstorage.resource_amount_input.no_input": "Enter amount",
+  "gui.refinedstorage.resource_amount_input.non_int": "Enter integer",
```
raoulvdberge, line 287: Check what can be removed here.

refinedstorage-query-parser/src/main/java/com/refinedmods/refinedstorage/query/evaluator/Evaluator.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+        this.nodes.addAll(parser.getNodes());
+    }
+
+    private List<Token> applyUnits(final List<Token> tokens) {
+        final List<Token> output = new ArrayList<>();
+        for (final Token token : tokens) {
+            if (token.type() == TokenType.IDENTIFIER) {
+                if (output.isEmpty()) {
+                    throw new EvaluatorException("resource_amount_input.no_value_before_unit");
```
raoulvdberge, line 71: Throw a custom exception instead of passing magical strings like this.

refinedstorage-query-parser/src/main/java/com/refinedmods/refinedstorage/query/evaluator/Evaluator.java
```diff
@@ -0,0 +1,124 @@
...
+                }
+                return Double.parseDouble(token.content());
+            }
+            case ParenNode(List<Node> binOpNode) -> {
+                return compute(binOpNode.getFirst());
+            }
+            default -> {
+            }
+        }
+        throw new EvaluatorException("Parser screwed up");
```
raoulvdberge:
```suggestion
        throw new EvaluatorException(String.format("Unhandled node %s", operand));
```

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -224,26 +234,22 @@ private void changeAmount(final int delta) {
...
+            correctedDelta,
             configuration.getMinAmount(),
             configuration.getMaxAmount()
-        )));
+        );
+        updateAmount(newAmount);
     }
 
-    private int correctDelta(final N oldAmount, final int delta) {
+    private int correctDelta(@Nullable final N oldAmount, final int delta) {
```
raoulvdberge, line 247: I don't understand why this is nullable now?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
```diff
@@ -224,26 +234,22 @@ private void changeAmount(final int delta) {
...
-                oldAmount,
-                correctedDelta,
-                configuration.getMinAmount(),
-                configuration.getMaxAmount()
-            );
-            updateAmount(newAmount);
-        }, () -> updateAmount(amountOperations.changeAmount(
-            null,
-            delta,
+        final int correctedDelta = correctDelta(getValue(), delta);
```
raoulvdberge, line 237: Where is getValue() coming from? This used to be optional with `getAndValidateAmount`?

raoulvdberge (changes requested):

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: What do you mean by resize the screen?

refinedstorage-query-parser/src/main/java/com/refinedmods/refinedstorage/query/evaluator/Evaluator.java
lucasser: Is that not what i'm doing? I have an EvaluatorException and i'm setting a message for it.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: To work with getValue which returns a Nullable

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: getValue is a wrapper for checkValue which returns a Nullable N instead of  AmountOperations.ReturnValue<N> So that you can deal with it in a couple of lines and not have to check for erroring.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: should i pass different error states with an enum instead?

lucasser: or like autocrafing error states are handled?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: Resize your minecraft window

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: Well again, I don't like that we have error states to begin with.

refinedstorage-query-parser/src/main/java/com/refinedmods/refinedstorage/query/evaluator/Evaluator.java
raoulvdberge: Those are magical strings, not actual error messages.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: Can't we just not call this method if it's null?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
raoulvdberge: Don't obfuscate code like that, just take the hit for those extra lines.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/support/amount/AbstractAmountScreen.java
lucasser: I'm trying to tell the user why their expression isn't compiling or evaluating. Should I be doing that, and if so, if there a better way? Or just have it highlight red.

raoulvdberge: I'd just highlight it red. First do the minimal thing. If it's not obvious, we will see this based on discord support requests or bugs. My assumption for now is that it's going to be obvious. We'll see if that is correct, but for now lets keep the code more minimal.

lucasser: then im going to make another branch, so i keep the error handling code if we want to use it later

lucasser: You think you could do some kind of survey, just out of interest to see if people prefer tooltip or no tooltip?

raoulvdberge: You can do it, I won't. I prefer to just drop it in the world and see what happens.

lucasser:
PR'ing another branch without error management. If people complain, i should be able to add it fairly easily.

PR #916, opened 2025-03-31
fix: portable grid and controller incorrectly setting block energy

### LINKED ISSUE #896: Portable Grid restores energy when placing as a block
Because it's setting the energy twice (once from the item and once from the custom block entity data)

lucasser:
fix #896 as well as controller which had the same behaviour


raoulvdberge:
Yes, but I think this will break when you use dismantling mode (which uses block entity data)

raoulvdberge:
Hey @lucasser, thanks for researching this!

Yes, indeed, it appears like the `passEnergyToBlockEntity` method is no longer needed. 
`ItemBlockEnergyStorage` is already setting custom block entity data when `stored` changes, and block entity data is applied when placing the block. So we do not need to transfer it twice.

Can you rebase this PR, cleanup, and remove the `passEnergyToBlockEntity` method?

Please, also put an explanation about this in the commit message for later reference.

Issue #976, opened 2025-04-26
Add a registry for registering custom ExternalStorageProviders


lucasser:
### Describe your enhancement

Add a registry for registering custom ExternalStorageProviders for the external storage based on the block it is placed on. Like the interface works currently. This will allow integrating with inventories like QIO, or an ae2 system in a much cleaner way. The external storage, when creating a new provider would check if the block entity in front of it is in the registry, if it is, then it'll do the custom provider, if not it'll fall back on the generic one.
(I am also willing to implement this)

raoulvdberge:
Is this needed? You can register a custom `ExternalStorageProviderFactory` if you'd want compatibility with QIO or AE2. RS loads it in within a `CompositeExternalStorageProvider`.

Issue #886, opened 2025-03-06
Shared Inventory + External storage = Item flickering in grid


Demsli:
https://github.com/user-attachments/assets/96c1cc79-9ecd-4b9b-982c-4f8c2d8b23e0

### Describe the bug

When connecting an external storage to an ender chest from EnderStorage or Dimensional Chest from DimStorage - items start to flicker in the opened grid. 

### How can we reproduce this bug or crash?

1. Place an ender chest from EnderStorage or Dimensional Chest from DimStorage and connect it to external storage
2. Insert any item into the chest inventory
3. Open grid
4. See items flickering

### What Minecraft version is this happening on?

Minecraft 1.21.1

### What NeoForge or Fabric version is this happening on?

21.1.126

### What Refined Storage version is this happening on?

refinedstorage-neoforge-2.0.0-milestone.4.14

### Relevant log output

```shell
This is what's shown in debug.log:
[06mar.2025 00:40:34.188] [Server thread/DEBUG] [com.refinedmods.refinedstorage.common.storage.externalstorage.AbstractExternalStorageBlockEntity/]: Loading storage for external storage with direction north @ BlockPos{x=-9, y=-60, z=-3}
[06mar.2025 00:40:34.188] [Server thread/DEBUG] [com.refinedmods.refinedstorage.common.grid.AbstractGridContainerMenu/]: com.refinedmods.refinedstorage.common.grid.CraftingGridContainerMenu@3b8cc0e2 received a change of -64 for ItemResource[item=minecraft:jungle_log, components={}]
[06mar.2025 00:40:34.193] [Render thread/DEBUG] [com.refinedmods.refinedstorage.common.grid.AbstractGridContainerMenu/]: ItemResource[item=minecraft:jungle_log, components={}] got updated with -64
[06mar.2025 00:40:34.194] [Render thread/DEBUG] [com.refinedmods.refinedstorage.api.grid.view.GridViewImpl/]: ItemResource[item=minecraft:jungle_log, components={}] was already found in the view list
[06mar.2025 00:40:34.194] [Render thread/DEBUG] [com.refinedmods.refinedstorage.api.grid.view.GridViewImpl/]: Actually updating ItemResource[item=minecraft:jungle_log, components={}] resource in the view list
[...]

PR #979, opened 2025-04-27
Registry for custom behaviour with external storage for specific blocks

lucasser:
Implements #976

### DIFF
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApi.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApi.java
@@ -95,6 +95,10 @@ public interface RefinedStorageApi {
 
     Collection<ExternalStorageProviderFactory> getExternalStorageProviderFactories();
 
+    void addExternalStorageProviderBlockFactory(ExternalStorageProviderFactory factory, Identifier blockId);
+
+    PlatformRegistry<ExternalStorageProviderFactory> getExternalStorageProviderBlocks();
+
     Collection<DestructorStrategyFactory> getDestructorStrategyFactories();
 
     void addDestructorStrategyFactory(DestructorStrategyFactory factory);
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/externalstorage/ExternalStorageBlock.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/externalstorage/ExternalStorageBlock.java
@@ -4,7 +4,6 @@ import com.refinedmods.refinedstorage.common.content.BlockColorMap;
 import com.refinedmods.refinedstorage.common.content.BlockEntities;
 import com.refinedmods.refinedstorage.common.content.BlockEntityProvider;
 import com.refinedmods.refinedstorage.common.content.Blocks;
-import com.refinedmods.refinedstorage.common.iface.InterfaceBlock;
 import com.refinedmods.refinedstorage.common.support.AbstractBlockEntityTicker;
 import com.refinedmods.refinedstorage.common.support.AbstractDirectionalCableBlock;
 import com.refinedmods.refinedstorage.common.support.BaseBlockItem;
@@ -85,9 +84,8 @@ public class ExternalStorageBlock extends AbstractDirectionalCableBlock
         super.neighborChanged(state, level, pos, block, orientation, movedByPiston);
         if (level instanceof ServerLevel serverLevel
             && level.getBlockEntity(pos) instanceof AbstractExternalStorageBlockEntity blockEntity) {
-            final boolean didBreakInterfaceBlock = block instanceof InterfaceBlock;
-            final boolean didPlacePotentialInterfaceBlock = block instanceof AirBlock;
-            if (didBreakInterfaceBlock || didPlacePotentialInterfaceBlock) {
+            final boolean placedSomething = block instanceof AirBlock;
+            if (placedSomething) {
                 blockEntity.loadStorage(serverLevel);
             } else {
                 blockEntity.neighborChanged();
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApiProxy.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApiProxy.java
@@ -129,6 +129,17 @@ public class RefinedStorageApiProxy implements RefinedStorageApi {
     }
 
     @Override
+    public void addExternalStorageProviderBlockFactory(final ExternalStorageProviderFactory factory,
+                                                       final Identifier blockId) {
+        ensureLoaded().addExternalStorageProviderBlockFactory(factory, blockId);
+    }
+
+    @Override
+    public PlatformRegistry<ExternalStorageProviderFactory> getExternalStorageProviderBlocks() {
+        return ensureLoaded().getExternalStorageProviderBlocks();
+    }
+
+    @Override
     public Collection<DestructorStrategyFactory> getDestructorStrategyFactories() {
         return ensureLoaded().getDestructorStrategyFactories();
     }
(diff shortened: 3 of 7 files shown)


Nodrance:
based

lucasser:
Will let us integrate with QIO and AE2 very easily.

Nodrance:
"is rs2 or ae2 better" we're doing both now keep up

raoulvdberge:
I'm not sure that this will work properly. The external storage API is designed to just be active for all providers all the time, regardless of block type. We can't be block type dependant. The reason for this is: if we are block type dependant, that means you need to reload the storage if the connected block changes (this currently doesn't happen, by design). If you reload when the connected block changes, this causes 2 problems:

* Items are flickering in the Grid
* Server strain

Listening to connected block changes isn't bad per se, but some blocks constantly send block updates. See #886 
And yes, indeed, interfaces get an exception for the neighbor changed event and do cause a reload.

Regarding possible AE integration:

Refined Storage shouldn't have integration with Applied Energistics, there is no use case for it and encourages people to create weird setups for no reason that create problems.

raoulvdberge:
If we were to merge this, only for QIO integration, `ExternalStorageBlock#neighborChanged` needs to be updated as well for block change detection.

lucasser:
I have two solutions. I changed the neighbourChanged method to re initialize if the changed block is in the registry. This might cause problems if the special block always sends updates. The other option is we store the block the storage is currently sitting on, and if it changes:
```java
if (savedBlock.getLootTable().registry() != newBlock.getLootTable().registry()) {
    blockEntity.loadStorage(serverLevel);
} else...
```
then we change the provider.

lucasser:
I am trying to prevent hardcoded blocks/values so that if there is something else that need a custom provider, it'll be easy to integrate (for example create storage system)

raoulvdberge:
> I have two solutions. I changed the neighbourChanged method to re initialize if the changed block is in the registry. This might cause problems if the special block always sends updates.

Yes, but that is less of a problem since we explicitly whitelist these blocks and can check them.

> The other option is we store the block the storage is currently sitting on, and if it changes:

That is bound to cause issues: you have to deal with syncing world state & cache invalidation.

raoulvdberge:
`brokeSpecialBlock` should probably check `isPresent` instead? Please thoroughly test these changes!

I do not want to create more instability in a phase of development where we already have so many bugs.

lucasser:
Also, this doesn't have to be an urgent thing, I'm just submitting PRs as I finish working on them for you to review whenever you can.

lucasser:
Is there a reason for updating external storage provider when an interface/special block is broken as opposed to just update when something is placed

lucasser:
I updated this PR to 2.0.0, and stress tested. It correctly updates the provider on world load, when a block gets placed, and when a block gets instant replaced in creative mode. The one thing that doesn't work is entangled blocks, since they will load the composite provider when you place it and not update the provider depending on what's inside (but they won't work with an interface either). To fix it there would have to be an EntangledBlockExternalStorageProvider which would correctly update it. Everything else works fine.

raoulvdberge:
Please rebase on latest `develop`

lucasser:
Done

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApi.java
```diff
@@ -95,6 +95,10 @@ public interface RefinedStorageApi {
 
     Collection<ExternalStorageProviderFactory> getExternalStorageProviderFactories();
 
+    void addExternalStorageProviderBlockFactory(ExternalStorageProviderFactory factory, Identifier blockId);
```
raoulvdberge, line 98: The class is called ExternalStorageProviderFactory, so the method should be called `addExternalStorageProviderFactory`.

raoulvdberge, line 98: Instead of taking in an `Identifier` I'd prefer to take in a `ResourceKey<Block>`, is that possible?

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/externalstorage/ExternalStorageBlock.java
```diff
@@ -85,9 +84,8 @@ protected void neighborChanged(final BlockState state, final Level level, final
         super.neighborChanged(state, level, pos, block, orientation, movedByPiston);
         if (level instanceof ServerLevel serverLevel
             && level.getBlockEntity(pos) instanceof AbstractExternalStorageBlockEntity blockEntity) {
-            final boolean didBreakInterfaceBlock = block instanceof InterfaceBlock;
-            final boolean didPlacePotentialInterfaceBlock = block instanceof AirBlock;
-            if (didBreakInterfaceBlock || didPlacePotentialInterfaceBlock) {
+            final boolean placedSomething = block instanceof AirBlock;
+            if (placedSomething) {
```
raoulvdberge, line 88: So why is the check for `didBreakInterfaceBlock` gone without alternative?

refinedstorage-fabric/src/main/java/com/refinedmods/refinedstorage/fabric/ModInitializerImpl.java
```diff
@@ -321,6 +324,12 @@ private void registerExternalStorageProviderFactories() {
         );
     }
 
+    private void registerCustomExternalStorageProviderFactories() {
```
raoulvdberge, line 327: Shouldn't this logic be shared in `AbstractModInitializer`?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+-   You can now register custom ExternalStorageProviders for specific blocks by calling addExternalStorageProviderBlockFactory inside registerCustomExternalStorageProviderFactories in the ModInitializer class
```
raoulvdberge, line 12: This is totally not relevant for the changelog, players don't care about this at all.

raoulvdberge (changes requested):

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApi.java
lucasser: There are two different `externalStorageProvider` groups.
```java
    private final List<ExternalStorageProviderFactory> externalStorageProviderFactories = new ArrayList<>();
    private final PlatformRegistry<ExternalStorageProviderFactory> customExternalStorageProviderBlocks =
```
one is for generic resource-based providers: item storage, fluid storage (chemical storage) and is an array
The other is for specific blocks (interface, QIO dash) and is a registry which can be queried for block id. Both of them consist of `ExternalStorageProviderFactories` but they are sepparate things. I could try to combine them, but i dont think that will be very helpful, or I can rename the methods/collections to `resourceExternalStorageProviderFactories` and `blockExternalStorageProviderFactories`, or i can add a class for block provider factories only.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/externalstorage/ExternalStorageBlock.java
lucasser: We don't need it. The only time the provider might have to change is when the targeted block gets changed. If it gets broken, that's irrelevant because air is not targeted either way. So the only time we need to reload the provider is if a block is placed. I tested different ways of changing blocks (piston, creative mode replacing) and all of them will first put an air block there, and then add the new block.
Tho only time this will not work with is with entangled blocks, but there could be an addon making an `entangledBlockProvider` to fix that.

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/RefinedStorageApi.java
lucasser: Right now the providers are stored in a `PlatformRegistry`, so no. But I can change that. Is there an example registry in `refinedStorageAPI` that I can ~copy~ follow?

lucasser: If you mean the function takes a ResourceKey and then converts it into an Identifier, then i can do that

raoulvdberge: Just stop storing these block providers in a registry?

lucasser: And store them in a map instead?

raoulvdberge:
Closing this for my question outlined in #976

PR #995, opened 2025-05-04
Fix #952: Preserve order of resources in PatternProviderSink

### LINKED ISSUE #952: Autocrafter ignoring pattern order
### Describe the bug

Making a pattern that goes into a machine where the order matters,
It looks like it inserts the items in alphabetical order and not pattern order

This is the pattern:
Image
This is the result:
Image

1.  Make pattern
2. Add to autocrafter facing an inventory
3. Look in inventory output
4. Items are in wrong order.

Minecraft 1.21.1

21.1.145

2.0.0-beta.2

### Relevant log output

```shell

```

Noobidoo:
Ensure that the order of resources is maintained when retrieving keys and copying state in the PatternProviderSink. This change modifies the way resources are collected to respect their original order.


raoulvdberge:
Why close this PR?

PR #1005, opened 2025-05-14
Fix/network node empty exception

MarkSG93:


raoulvdberge:
What is this supposed to be?

PR #1058, opened 2025-07-18
Update PortableGridBlockItem.java

DatJaneDoe:
When writing the new compound tag for the Portable Grid when changing the disk inventory, include the current energy storage so it doesn't reset to 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [2.0.0-beta.3] - 2025-07-22
+
 ### Changed
 
 -   Closing the Crafting Preview window (via cancel or escape) will now properly cancel pending calculations.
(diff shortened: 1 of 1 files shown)


raoulvdberge:
What is the bug here?

raoulvdberge:
Please describe it in terms of functionality, not in terms of code.

DatJaneDoe:
> What is the bug here?

The Portable Grid's energy resets to zero when the disk is changed. This commit fixes the bug.

DatJaneDoe:
> Please describe it in terms of functionality, not in terms of code.

In addition to the bug fix, I reworked Portable Grids placement code to be more practical like backpacks. (Mostly copied from my Omni Grid addon)

-Right clicking with the Portable Grid always opens the GUI. 
-Sneak-right-clicking always attempts to place it.
-You can pick up the Portable Grid with shift-right-click with an empty main hand.

raoulvdberge:
That's not how I prefer to work. No stacking unrelated changes in the same PR, no merge commits on a PR branch, etc (check the failed workflows)

PR #1105, opened 2025-08-10
fix: show correct drain amount on hover

starforcraft:
When hovering over a fluid tank that can extract more than one bucket of fluid, the tooltip displays a different amount than what is actually drained when right-clicking. Regardless of the tooltip value, right-clicking always drains exactly one full bucket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 -   Duplication bug with Inventory Essentials.
+-   Incorrect extractable amount shown when hovering over a fluid tank in the Grid.
 
 ## [2.0.0-beta.9] - 2025-08-15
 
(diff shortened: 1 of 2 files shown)


raoulvdberge:
This breaks a few things:

- Storage monitor will now only insert 1 bucket at a time
- Putting a fluid container in a filter slot will configure 1 bucket instead of the amount stored in the tank

starforcraft:
I’ve updated it so the tooltip only changes the displayed amount now

raoulvdberge:
This is great, thanks. Please update the changelog to put the entry under the "fixed" section and I'll merge this.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+-   Fixed incorrect extractable amount shown when hovering over a fluid tank
```
raoulvdberge:
```suggestion
-   Fixed incorrect extractable amount shown when hovering over a fluid tank in the Grid.
```

raoulvdberge (changes requested):

raoulvdberge (approved):

PR #1118, opened 2025-08-19
Reword the old pull request i made.

FijiAura:
I was trying to say that you can just install both mods at once.


raoulvdberge:
I don't think this is an important distinction to make.

PR #1125, opened 2025-08-21
perf: resolve item tooltip attribute lazily

SirYwell:
This PR at least partially addresses slow Grid opening.
I noticed that a lot of time is spent in `AbstractItemGridResourceRepositoryMapper#getTooltip(ItemStack)`. However, this information is only ever used when searching. It shouldn't slow down opening the grid.

By only resolving the tooltip lazily, the expensive part only happens when it is actually needed.

I backported this change to beta.3 as this is what's currently running on my ATM10 server. There, `ViewList#createSorted(...)` previously took 1-2s each time when opening. With this change, it's down to 50-100ms. I measured using a custom JFR event and repeatedly open/close the grid to ensure the JIT kicks in first.

Please let me know what you think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+-   Reduce lag spikes when opening a Grid with many item types. 
+
 ### Added
 
 -   Warning to the resource type side button when the Grid is filtering by resource type and there are no resources for that type to display.
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
@@ -17,11 +17,20 @@ import org.apiguardian.api.API;
 public abstract class AbstractGridResource<T extends PlatformResourceKey> implements GridResource {
     protected final T resource;
     private final String name;
-    private final Map<GridResourceAttributeKey, Set<String>> attributes;
+    private final Function<GridResourceAttributeKey, Set<String>> attributes;
 
+    // TODO: merge constructors for >=1.22
     protected AbstractGridResource(final T resource,
                                    final String name,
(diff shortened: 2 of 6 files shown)


raoulvdberge:
Thank you, this is really great.
I thought slow grid opening was mainly due to network overhead for the initial packet, with lots of items.
But in your analysis it seems to be due to slow tooltip providers.

SirYwell:
Yes, it's at least the largest contributor when creating the menu. I have spark reports of just opening and closing the inventory for 11 seconds (because that's spark's minimum timeout):
- Before this change: https://spark.lucko.me/T0mOFVJnl0
- With this change: https://spark.lucko.me/fIkdE4wk4p

These aren't 100% comparable as I'm just a human, but focusing on the `BlockableEventLoop#runAllTasks()` method, I think that's comparable enough.

raoulvdberge:
Can you add a changelog entry and rebase (not merge) on develop? I broke the build on there but fixed it now.

raoulvdberge:
Oh, one more thing. Could you add an alternate constructor that still accepts `Set<String>` in `AbstractGridResource`. I don't want to break addons at this point.

And can you change it so item tags are also lazy resolved? 

Thanks

raoulvdberge:
Checkstyle appears to be failing, make sure to import the intellij coding style from `config/`

SirYwell:
> Oh, one more thing. Could you add an alternate constructor that still accepts `Set<String>` in `AbstractGridResource`. I don't want to break addons at this point.

Such constructor would clash due to type erasure. If this is important to you, maybe it makes sense to use something else than `Map` (the way it is currently used, `Function<? super GridResourceAttributeKey, ? extends Set<String>>` would work, or potentially a custom type). WDYT?

raoulvdberge:
Yes, we could do a temporary workaround and use something else than `Map`. Don't have any preference. Could add a TODO that we go to 1 canonical constructor in RS 3 (for MC 1.22).

(Normally I would just break the ctor for the beta releases but modpack devs aren't happy if I break addons)

SirYwell:
I updated the code to ensure backwards compatibility and checkstyle should be happy now too.

raoulvdberge:
Thanks. Could you bring the branch up to date by rebasing?

SirYwell:
> Thanks. Could you bring the branch up to date by rebasing?

Done!

Issue #956, opened 2025-04-20
CraftingTree recursion calling MutableResourceListImpl.copy on nested crafts causing lag


SirTremain:
### Describe your enhancement

On the server I am running, I noticed a lot of lag coming from some of the refined storage exporters with autocrafting upgrades. I ran observable sampler over it and found that the exporters call AutocraftingNetworkComponentImpl.getMaxAmountSync, which leads to the CraftingTree object simulating the craft to see how many it can craft. At every crafting level, the CraftingTree calls child() for all it's children and creates a copy of the current Crafting state to simulate the child crafts over. Following the sampler I found that the calls to MutableResourceListImpl.copy make up a majority of the CPU time.

MutableResourceListImpl.copy calls the it's add() method on each entry which in turns attempts to get the objects from it's entries. This causes many hashes to be created for objects despite them definitely not being contained inside the crafting state at this point. As the crafting state starts empty, every object will be put into the craftingState for the first time. This calls the addNew method anyway as it's the first time it sees this entry.

This is a screenshot from the profiler. This contains only one branch of the tree, but they all terminate the same way, with the MutableResourceListImpl.copy being most of the CPU time:

![Image](https://github.com/user-attachments/assets/0b161c33-afa2-46a8-8d46-29113d611f5f)

Can I suggest changing this .add call in MutableResourceListImpl.copy to directly call addNew instead? It seems it is eventually calling addNew anyway so it should only reduce the number of hashMap.get calls. Obviously I'm not an expert so please let me know if I'm missing anything. 

Extra: Ideally copying the hashMap could be done in a different way entirely. I'm not super familiar with java but in rust for instance it would be possible to instead to copy the memory directly vs adding each new entry one a time. Maybe it would be possible in java using serialization/deserialization to copy the object? Since the copy should be identical, anything to avoid creating the hashes again would go a long way.

raoulvdberge:
Mostly fixed by ea8fce73af9ae4c3994593372097d791028233a4.

Issue #1100, opened 2025-08-09
Auto-crafter is no longer able to calculate items


ZEkA10000:
### Describe the bug

Earlier I used Refined Storage 2.0.0-beta3 (Neo Forge 1.21.1) and it calculated all resources for crafting 16384k storage part (Extra Storage add-on), but yesterday it showed, "Process is too long, calculation cancelled". It also happened even for 64k part (creating from nothing) sometimes.

I used Diamond Auto-Crafter (Extra add-on)

<img width="981" height="422" alt="Image" src="https://github.com/user-attachments/assets/18c8f085-5110-4b82-9289-530b4602bb9c" />

### How can we reproduce this bug or crash?

1. Build a storage
2. Add auto-crafter
3. Request auto-craft

### What Minecraft version is this happening on?

Minecraft 1.21.1

### What NeoForge or Fabric version is this happening on?

Neoforge 1.21.1 v21.1.193

### What Refined Storage version is this happening on?

2.0.0-beta3 and 2.0.0-beta8

### What modpack and modpack version, if any, are you using?

Custom

### Relevant log output

```shell

```

raoulvdberge:
Yes, Refined Storage has some difficulty with large recipes right now.
Read my comment at https://github.com/refinedmods/refinedstorage2/issues/980#issuecomment-3102627403

alexandrusendroni:
Try to craft the components. One of them has missing materials.

ZEkA10000:
> Try to craft the components. One of them has missing materials.

Idk why,  but when I installed EMI, autocrafter calculated 16m storage cell from scratch. Without that mod, I had that error.

raoulvdberge:
In beta 12 this should be less an issue if you have all the materials at least (by #1129)
Closing this issue for now and let me know if it improved at all.

raoulvdberge:
Leaving this one open because there are still some improvements we can do, mentioned in the PR #1129

KindaWorks:
Could we have a config value with an amount of how many layers deep autocrafting would go? Increasing it could result in longer wait times, but it would be a conscious decision of an admin.

raoulvdberge:
Did you try out the latest beta and check if it's still an issue?

kfp:
In my initial testing the beta.13 cleared up the issue I was having with some particularly complex Modern Industrialization crafts timing out. Thanks!

raoulvdberge:
Awesome!

[...]

PR #1129, opened 2025-08-23
perf: optimize autocrafting calculation

SirYwell:
Also related to #1100, but while this change provides significant speedups, timeouts are still common when materials are missing.

One main observation is that we most likely don't touch most of the material in the system. Consequently, if we track the changes on top of the original system state, we can reduce copies to what the actual changes.

On top of that, we can avoid HashMap resizing by pre-sizing them better and we can also avoid lookups when copying as described in #956.

In my testings, even complex recipes get calculated when enough material is present. With material is missing, it is also a lot faster, but noticeably slower.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Fixed
 
 -   Rare crash when starting an autocrafting task.
+-   Optimize autocrafting calculation performance.
 
 ## [2.0.0-beta.11] - 2025-08-25
 
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingState.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingState.java
@@ -8,17 +8,13 @@ import com.refinedmods.refinedstorage.api.resource.list.MutableResourceList;
 import com.refinedmods.refinedstorage.api.resource.list.MutableResourceListImpl;
 import com.refinedmods.refinedstorage.api.storage.root.RootStorage;
 
-import java.util.Comparator;
-
 class CraftingState {
     private final MutableResourceList storage;
     private final MutableResourceList internalStorage;
-    private final Comparator<ResourceKey> sorter;
 
     private CraftingState(final MutableResourceList storage, final MutableResourceList internalStorage) {
         this.storage = storage;
         this.internalStorage = internalStorage;
-        this.sorter = createSorter(storage).thenComparing(createSorter(internalStorage));
     }
 
     void extractFromInternalStorage(final ResourceKey resource, final long amount) {
@@ -51,29 +47,28 @@ class CraftingState {
         return new CraftingState(LazyCopyMutableResourceList.create(storage), MutableResourceListImpl.create());
     }
 
-    Comparator<ResourceKey> getSorter() {
-        return sorter;
-    }
-
-    private static Comparator<ResourceKey> createSorter(final MutableResourceList list) {
-        return (a, b) -> {
-            final long ar = list.get(a);
-            final long br = list.get(b);
-            return (int) br - (int) ar;
-        };
-    }
-
     ResourceState getResource(final ResourceKey resource) {
         return new ResourceState(resource, storage.get(resource), internalStorage.get(resource));
     }
 
-    record ResourceState(ResourceKey resource, long inStorage, long inInternalStorage) {
+    record ResourceState(ResourceKey resource, long inStorage, long inInternalStorage)
+        implements Comparable<ResourceState> {
         boolean isInStorage() {
             return inStorage > 0;
         }
 
         boolean isInInternalStorage() {
             return inInternalStorage > 0;
+        }
+
+        @Override
+        public int compareTo(final ResourceState o) {
+            // o first, we want greater values to come first
+            final int storage = Long.compare(o.inStorage, inStorage);
+            if (storage == 0) {
+                return Long.compare(o.inInternalStorage, inInternalStorage);
+            }
+            return storage;
         }
     }
 }
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingState.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingState.java
@@ -3,6 +3,7 @@ package com.refinedmods.refinedstorage.api.autocrafting.calculation;
 import com.refinedmods.refinedstorage.api.autocrafting.Pattern;
 import com.refinedmods.refinedstorage.api.resource.ResourceAmount;
 import com.refinedmods.refinedstorage.api.resource.ResourceKey;
+import com.refinedmods.refinedstorage.api.resource.list.LazyCopyMutableResourceList;
 import com.refinedmods.refinedstorage.api.resource.list.MutableResourceList;
 import com.refinedmods.refinedstorage.api.resource.list.MutableResourceListImpl;
 import com.refinedmods.refinedstorage.api.storage.root.RootStorage;
@@ -47,7 +48,7 @@ class CraftingState {
     static CraftingState of(final RootStorage rootStorage) {
         final MutableResourceListImpl storage = MutableResourceListImpl.create();
         rootStorage.getAll().forEach(storage::add);
-        return new CraftingState(storage, MutableResourceListImpl.create());
+        return new CraftingState(LazyCopyMutableResourceList.create(storage), MutableResourceListImpl.create());
     }
 
     Comparator<ResourceKey> getSorter() {
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingTree.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingTree.java
@@ -87,7 +87,7 @@ class CraftingTree<T> {
     private CalculationResult calculateIngredient(final int ingredientIndex, final IngredientState ingredientState,
                                                   final CancellationToken cancellationToken)
         throws CancellationException {
-        CraftingState.ResourceState resourceState = craftingState.getResource(ingredientState.get());
+        CraftingState.ResourceState resourceState = ingredientState.get();
         long remaining = ingredientState.amount() * amount.iterations();
         if (remaining < 0) {
             throw new NumberOverflowDuringCalculationException();
@@ -136,7 +136,6 @@ class CraftingTree<T> {
             return calculateChild(ingredientState, remaining, childPatterns, resourceState, cancellationToken);
         }
         return ingredientState.cycle()
-            .map(craftingState::getResource)
             .orElseGet(() -> {
                 listener.ingredientsExhausted(resourceState.resource(), remaining);
                 return null;
@@ -192,7 +191,7 @@ class CraftingTree<T> {
     @Nullable
     private CraftingState.ResourceState cycleToNextIngredientOrFail(final IngredientState ingredientState,
                                                                     final ChildCalculationResult<T> childResult) {
-        return ingredientState.cycle().map(craftingState::getResource).orElseGet(() -> {
+        return ingredientState.cycle().orElseGet(() -> {
             this.craftingState = childResult.childTree.craftingState;
             listener.childCalculationCompleted(childResult.childTree.listener);
             return null;
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/IngredientState.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/IngredientState.java
@@ -8,20 +8,20 @@ import java.util.Optional;
 
 class IngredientState {
     private final long amount;
-    private final ResourceKey[] possibilities;
+    private final CraftingState.ResourceState[] possibilities;
     private int pos;
 
     IngredientState(final Ingredient ingredient, final CraftingState state) {
         this.amount = ingredient.amount();
-        this.possibilities = new ResourceKey[ingredient.inputs().size()];
+        this.possibilities = new CraftingState.ResourceState[ingredient.inputs().size()];
         for (int i = 0; i < ingredient.inputs().size(); i++) {
             final ResourceKey resource = ingredient.inputs().get(i);
-            possibilities[i] = resource;
+            possibilities[i] = state.getResource(resource);
         }
-        Arrays.sort(possibilities, state.getSorter());
+        Arrays.sort(possibilities);
     }
 
-    ResourceKey get() {
+    CraftingState.ResourceState get() {
         return possibilities[pos];
     }
 
@@ -29,11 +29,11 @@ class IngredientState {
         return amount;
     }
 
-    Optional<ResourceKey> cycle() {
+    Optional<CraftingState.ResourceState> cycle() {
         if (pos + 1 >= possibilities.length) {
             return Optional.empty();
         }
         pos++;
-        return Optional.of(get());
+        return Optional.of(possibilities[pos]);
     }
 }
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/preview/PreviewBuilder.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/preview/PreviewBuilder.java
@@ -11,11 +11,16 @@ import java.util.List;
 import java.util.Map;
 
 public class PreviewBuilder {
-    private final Map<ResourceKey, MutablePreviewItem> items = new LinkedHashMap<>();
+    private final Map<ResourceKey, MutablePreviewItem> items;
     private List<ResourceAmount> outputsOfPatternWithCycle = Collections.emptyList();
     private boolean missing;
 
     private PreviewBuilder() {
+        items = new LinkedHashMap<>();
+    }
+
+    private PreviewBuilder(final int size) {
+        items = LinkedHashMap.newLinkedHashMap(size);
     }
 
     public static PreviewBuilder create() {
@@ -51,7 +56,7 @@ public class PreviewBuilder {
     }
 
     public PreviewBuilder copy() {
-        final PreviewBuilder copy = new PreviewBuilder();
+        final PreviewBuilder copy = new PreviewBuilder(items.size());
         for (final Map.Entry<ResourceKey, MutablePreviewItem> entry : items.entrySet()) {
             final MutablePreviewItem item = entry.getValue();
             copy.items.put(entry.getKey(), item.copy());
--- a/refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceList.java
+++ b/refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceList.java
@@ -1,0 +1,112 @@
+package com.refinedmods.refinedstorage.api.resource.list;
+
+import com.refinedmods.refinedstorage.api.resource.ResourceAmount;
+import com.refinedmods.refinedstorage.api.resource.ResourceKey;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+import org.apiguardian.api.API;
+
+/**
+ * An implementation of a {@link ResourceList} that lazily copies from an original list.
+ * Only the entries that are changed are copied.
+ */
+@API(status = API.Status.STABLE, since = "2.0.0-beta.11")
+public class LazyCopyMutableResourceList implements MutableResourceList {
+    private final MutableResourceList original;
+    private final MutableResourceList updates;
+    private final Set<ResourceKey> pulled;
+
+    private LazyCopyMutableResourceList(final MutableResourceList original, final MutableResourceList updates,
+                                        final Set<ResourceKey> pulled) {
+        this.original = original;
+        this.updates = updates;
+        this.pulled = pulled;
+    }
+
+    public static LazyCopyMutableResourceList create(final MutableResourceList original) {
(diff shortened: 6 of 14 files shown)


refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceListImpl.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @Override
+    public void clear() {
+        pulled.addAll(original.getAll());
+        updates.clear();
+    }
+
+    @Override
+    public MutableResourceList copy() {
+        return new LazyCopyMutableResourceListImpl(original, updates.copy(), new HashSet<>(pulled));
```
SirYwell: The `new HashSet<>(pulled)` call shows up now when profiling. I think it is fine, but there might be alternative approaches, e.g., creating a fresh HashSet instead and passing updated as a new original (i.e., build a deeper tree instead of the flat design right now).
That would add some more complexity.

It might also make sense to explore other Set implementations, (e.g., from fastutil) and see if that improves things further.

refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceListImpl.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Set;
+import javax.annotation.Nullable;
+
+import org.apiguardian.api.API;
+
+/**
+ * An implementation of a {@link ResourceList} that lazily copies from an original list.
+ * Only the entries that are changed are copied.
+ */
+@API(status = API.Status.STABLE, since = "2.0.0-beta.11")
```
SirYwell: Not sure if this is the right

raoulvdberge:
This is interesting. I will check this with some complex recipes on All The Mods.

In the meantime, I wrote this test (https://gist.github.com/raoulvdberge/029a5d0650a1f3f9d573534fb1ef3565) a while ago when trying to test some optimizations. At the time, I was trying to optimize autocrafting by introducing threading. Was a failed attempt, only made things slower.

But the test is still good, as it simulates complex recipes with lots of substitutions.
What do you think of it? How does it perform on your branch vs `develop`? Is it realistic?

refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceListImpl.java
raoulvdberge: I think we will always have some hot spot that will show up in the profiler, I think this is fine. We can introduce more complexity later.

However, I am interested in how fastutil performs on some hotspots (like `ResourceList`). But that is something I'd ideally like to measure/compare in another PR.

refinedstorage-resource-api/src/main/java/com/refinedmods/refinedstorage/api/resource/list/LazyCopyMutableResourceListImpl.java
raoulvdberge: It's right, but you can remove the `Impl` from the name. It's not needed here, as the name of the implementation is different than the interface.

SirYwell:
> This is interesting. I will check this with some complex recipes on All The Mods.
> 
> In the meantime, I wrote this test ([gist.github.com/raoulvdberge/029a5d0650a1f3f9d573534fb1ef3565](https://gist.github.com/raoulvdberge/029a5d0650a1f3f9d573534fb1ef3565)) a while ago when trying to test some optimizations. At the time, I was trying to optimize autocrafting by introducing threading. Was a failed attempt, only made things slower.
> 
> But the test is still good, as it simulates complex recipes with lots of substitutions. What do you think of it? How does it perform on your branch vs `develop`? Is it realistic?

Interesting test. My changes seem to perform worse there. It looks like sorting in the `IngredientState` constructor is the main contributor, as we now have 2 map lookups instead of 1 per list access (and there are 2 list accesses per comparison).
But the test is missing one central point this PR addresses: There are only 4 resources in the RootStorage. On our server, we currently have >14k resources there. That means this test doesn't really reflect the cost of copying all these resources properly.

But the slowdown in this test also comes down to one central pain point: The slow hashCode calculation of `ItemResource`. Maybe it makes sense to convert that record to a normal class and cache the hashCode once calculated (I'm not sure how often these are re-created, but with this kind of double-lookup I'd expect to see an improvement). I'll experiment with that.

raoulvdberge:
Would you say the test is too extreme if it comes down to crafting complexity, and is not extreme enough if it comes down to root storage size?

Then I wonder, what is the main reason why autocrafting recipes can't calculate in time on a typical Minecraft server: crafting complexity or root storage size?

I was thinking that it's crafting complexity, but maybe that is a wrong assumption, and my test is testing the wrong component, so it doesn't really matter that your code performs worse.

SirYwell:
Hm, at least in our scenario (crafting extra disk 1024k storage parts with the mentioned root storage size) the copying seems to be a far larger contributor. But I'm not sure if this is considered a complex recipe already.

One observation regarding the sorting: Even if the array is already sorted by coincidence, we need to get the amount of each resource at least once to verify it is sorted. That means we could pre-fetch the amount before sorting, and sort using that. It looks like we could reuse `CraftingState.ResourceState` there. From a quick test, it seems like we can get even better performance this way:

<table>
<tr>
 <td>Root Storage Size
 <td>develop
 <td>local changes
<tr>
 <td>4
 <td>239s
 <td>178s
<tr>
 <td>4 + 1000
 <td>322s
 <td>206s
</table>

I also noticed that `Pattern#hashCode` takes ~50s in every scenario above. Given a Pattern has a UUID, I wonder if the hashCode could be simplified to return the hashCode of the UUID only? This would be an improvement orthogonal to this change.

SirYwell:
Another observation, and -- if I didn't miss any problems -- potentially the most relevant optimization: `PatternLayout` currently has an Ingredient list. For recipes where one ingredient appears in multiple slots (e.g., all the storage part recipes, where you need 3 parts of the previous tier), this means the list potentially contains the same *resource* multiple times. But that means crafting Item A from 3xB calculates 3 subtrees, while crafting 3xB directly only calculates one tree. *Especially* in these typical recipes where multiple lower tiers are used to craft higher tiers, this means we run into exponential search complexity, when linear complexity would be enough.

As an example, instead of the following search tree:

```mermaid
flowchart TD
    A[4k Storage Part]
    A -->|1| B1
    A -->|1| B2
    A -->|1| B3
    A -->|1| B4
    A -->|1| B5
    A -->|1| B6
    A -->|1| B7
    A -->|1| B8
    A -->|1| B9
    B1[1k Storage Part]
    B2[1k Storage Part]
    B3[1k Storage Part]
    B4[Basic Processor]
    B5[Basic Processor]
    B6[Basic Processor]
    B7[Basic Processor]
    B8[Redstone]
    B9[Iron Ingot]
    B1 -->|1| C1[Silicon]
    B1 -->|1| C2[Silicon]
    B1 -->|1| C3[Silicon]
    B1 -->|1| C4[Silicon]
    B1 -->|1| C5[Glass]
    B1 -->|1| C6[Glass]
    B1 -->|1| C7[Glass]
    B1 -->|1| C8[Iron Ingot]
    B1 -->|1| C9[Redstone]
```
    
 we can simplify the search to
 
 ```mermaid
flowchart TD
    A[4k Storage Part]
    A -->|3| B1
    A -->|4| B4
    A -->|1| B8
    A -->|1| B9
    B1[1k Storage Part]
    B4[Basic Processor]
    B8[Redstone]
    B9[Iron Ingot]
    B1 -->|3 * 4| C1[Silicon]
    B1 -->|3 * 3| C5[Glass]
    B1 -->|3* 1| C8[Iron Ingot]
    B1 -->|3 * 1| C9[Redstone]
 ```
 In my tests, this results in basically instant results even for complex recipes and also when resources are missing.

 Is there anything I'm missing that would prevent this optimization?

SirYwell:
I pushed my changes to the IngredientState possibilities sorting and PatterLayout Ingredient deduplication as separate commits so you can take a look. Please let me know what you think.

raoulvdberge:
I love it. Bit busy with work this week, I'll try to take a look as soon as I can. Thanks a lot so far for all the investigation.

raoulvdberge:
In the meantime, please check the test failures and checkstyle problems.
I think there were also some mutation testing problems in an earlier commit (this means that some new tests are missing, or existing tests need more assertions.)

SirYwell:
Looking into the test failures:

- `PreviewTest`, `TreePreviewTest` (`shouldPrioritizeResourcesThatWeHaveMostOfInStorageForSingleRootPatternAndMultipleIngredients()`): My fault! The `compare` implementation was wrong, good that these tests exist :)
- The other tests seem all to fail for the same reason, there is indeed different behavior now:
	-  Previously, when crafting 3 crafting tables from `4 x oak planks, 1 x oak log, 1 x spruce log`, we would for each of the 4 planks needed for a crafting table try to find 3 resources.
		1. Use 3 of the 4 oak planks
		2. Use the last oak plank and craft 4 more (=> 2 oak planks in the internal storage)
		3. Use the 2 oak planks and craft 4 spruce planks (=> 3 spruce planks in the internal storage)
		4. Use the remaining 3 spruce planks
	- With my most recent change, we look once for 12 planks instead.
		- With crafting we get to 8 oak planks, but that's not enough => MISSING_RESOURCES
		- With crafting we get to 4 spruce planks, but that's not enough => MISSING_RESOURCES
	- This seems odd, not sure if this is the intended behavior. Especially as we had mixed planks before already! I think MISSING_RESOURCES should only be the result if all possibilities together cannot satisfy the needs. But I wonder if crafting should be preferred over using multiple different resources in that case. If not, we would need to track craftable items and try them after all available resources were checked/consumed.

SirYwell:
I updated the commits. Regarding the mutation tests, I reworked some stuff to allow predefined lists (needed for the lazy copy one, as it's impossible to add to the original list later). Please let me know what you think of the changes there.

The sorting is fixed now, but the crafting issue remains for now. I'll wait for your input there before proceeding with anything.

raoulvdberge:
Ah yes, it is important that we keep the existing behavior. It needs to process these ingredient by ingredient in order to make sure that mixed ingredients still work; this is what makes RS2 autocrafting so powerful.

SirYwell:
> Ah yes, it is important that we keep the existing behavior. It needs to process these ingredient by ingredient in order to make sure that mixed ingredients still work; this is what makes RS2 autocrafting so powerful.

That's unfortunate, I removed the last patch (but I still got it here https://gist.github.com/SirYwell/857e2b6766698ac9b4633fcdc97a2be0 in case it gets relevant again somehow). I wonder what parts of the behavior need to be kept 1:1 and to what extent the behavior could be changed while keeping RS2 powerful. Maybe there are different ways to get similar speedups without violating these constraints. 

During this deep-dive I also noticed that there are cases where it produces unexpected results:

1. Create a fuzzy pattern for sticks
2. Create a pattern for oak fence gate
3. Put 2 x oak planks and 2 x other planks into the system
4. Try crafting an oak fence gate. Instead of using the other planks for the sticks, 1 oak plank is used for the sticks. Then, only 1 is available anymore but 2 are needed.

I assume resolving that would need a DAG instead of a tree, with some kind of dynamic reallocation of resources. But that sounds even more computation-heavy (but maybe allows for deduplication?). I'll try to do more research on that.

raoulvdberge:
Okay, some more time now. Like I mentioned I wanted to do some testing myself too.

I did some testing on All The Mods 10 crafting the **1048m Storage Disk** (fuzzy mode "on", on all patterns) and seeing promising results.

Analysing the table, it seems like this PR resolves the time-out problem for large crafts with no missing ingredients AND lots of unique items (before, it would time-out even if we had enough ingredients).

| Branch    | Missing Ingredients | Unique Items | Result     |
|-----------|---------------------|--------------|------------|
| develop   | Yes                 | 0          | Time-out  |
| develop   | No                  | 10          | No time-out  |
| develop   | Yes                 | 55633           |   (Didn't test, assuming this will also time-out given that missing ingredients with 0 unique items already times out) |
| develop   | No                  | 55633           | Time-out  |
| Your PR   | Yes                 | 0          |   Time-out  |
| Your PR   | No                  | 10          |  No time-out |
| Your PR   | Yes                 | 55633           | (Didn't test, assuming this will also give a time-out given that missing ingredients with 0 unique items already times out) |
| Your PR   | No                  | 55633           | No time-out  |

(This is consistent with your findings mentioned in the PR description)

> - 10 unique items: all the base ingredients necessary for the 1048m Storage Disk
> - 55633 unique items: all the items in the item registry EXCLUDING:
>   - Items from Extra Storage
>   - Items from Extra Disks
>   - Items from Refined Storage
>   - Glass (to ensure it uses the smelting recipe from Sand)
>   - Iron (to ensure it uses the smelting recipe from Iron Ingots)

My conclusion:

- No missing ingredients, and few unique items: never caused an issue
- Missing ingredients, few unique items: an issue on `develop` but also on this PR
- No missing ingredients, and lots of unique items: causes an issue on `develop`, and fixed by this PR
- Missing ingredients, and lots of unique items: an issue on `develop` but also on this PR

And a conclusion regarding the performance of your PR on my complex crafting test (please challenge me if incorrect):

- My complex crafting test is NOT realistic compared to complex recipes in Minecraft.
  - Why? The test completes in 30 sec., while complex crafts in Minecraft calculate in around 2 sec.
  - This indicates to me that the depth/complexity of the crafted resource in the test is too high.
- However, I do think it is a good benchmark for the autocrafting code.
- Right now it isolates crafting complexity. To be even more realistic, it would need to have a bigger root storage to also take the copying cost into account.

raoulvdberge:
Some questions I have for you:

- Could you rebase this PR on the latest `develop`? I'll merge this PR after that.
- Does this PR fix issue #956? (I'm assuming yes)
- Could you look into using fastutil in some hotspots like `ResourceList` and tell me if there is a measurable difference (in a separate PR)?
- Like you mentioned, how does improving the hashcode calculation of `ItemResource` and `Pattern` affect the performance?
- How can we avoid time-outs when there are missing ingredients but keep the existing behavior (or maybe a loosened variant of that)?
  - How far can we get with optimization fixes like this PR, without overhauling the entire autocrafting logic? 

Thanks for all the research you did so far. I'm really happy someone wants to help out with performance!

SirYwell:
> * Could you rebase this PR on the latest `develop`? I'll merge this PR after that.

Sure, will do so later.

> * Does this PR fix issue #956?

At least to some extent. If ingredients are missing, I'd expect the problem to be less extreme but still existing. But now the performance isn't directly coupled to the number of unique items anymore, so "cheap" recipes with high numbers of unique items should definitely see improvements. But more below regarding hashing.

> * Could you look into using fastutil in some hotspots like `ResourceList` and tell me if there is a measurable difference (in a separate PR)?

I can, but I'd first look into hash code calculation, that seems to be a bigger contributor right now.

> * Like you mentioned, how does improving the hashcode calculation of `ItemResource` and `Pattern` affect the performance?

I'd need to take new measurements there, but I have some motivating examples for `ItemResource`:
- Example of autocrafting using an exporter: https://spark.lucko.me/S70TBsnvL2 (look for `ItemResource.hashCode` in the flat view)
- Example of interactions via the Advanced Peripherals RS Bridge: https://spark.lucko.me/zI58wF9cOW?hl=1940 (though I think this also is sub-optimal usage of the API, as `getAll` would probably enough?)
- Example of a client lag when interacting with JEI: https://spark.lucko.me/j2qV5byfu9?hl=2329
- Example of `PreviewBuilder.copy` (I disabled the timeout there, the whole preview calculation was 16 seconds): 
preview-builder-copy

- In the same example, adding and removing the pattern to `activePatterns` in `CraftingTree` takes around 900ms. This is the only case where I saw it being somewhat relevant. So while it's not that much in sum, basically all of it is spent on hashing `PatternLayout`. If hashing the UUID is enough, that would be an extremely simple improvement.

> * How can we avoid time-outs when there are missing ingredients but keep the existing behavior (or maybe a loosened variant of that)?
> 
>     * How far can we get with optimization fixes like this PR, without overhauling the entire autocrafting logic?

Keeping the existing behavior exactly like that makes improvements that aren't just very local rather difficult. While it might be possible to squeeze out a bit more performance with micro-optimizations (e.g. replacing streams with loops), we're getting closer to the point where the limitation lies more in the algorithm than the implementation.

The change I experimented with could maybe get close to the current behavior with more adjustments. I'm not sure how much effort that would be though. Recall https://github.com/refinedmods/refinedstorage2/pull/1129#issuecomment-3220518111, a possible change would need to somehow keep track of partial results (i.e., first try to satisfy the requirement using *multiple* resources if multiple variants are allowed, not just one).

I'm not sure how well that would work with fuzzy processing patterns, when e.g., 2 stacks of logs are required: With this change, we would allow stuff like 1x oak log + 127 spruce log. Would it be a requirement to have 2 stacks of 1 material each in this case? 
Maybe there could be 2 paths, and deduplication just doesn't happen for processing patterns then.

SirYwell:
> * And a conclusion regarding the performance of your PR on my complex crafting test (please challenge me if incorrect):
> 
>     * My complex crafting test is NOT realistic compared to complex recipes in Minecraft.
>       
>       * Why? The test completes in 30 sec., while complex crafts in Minecraft calculate in around 2 sec.
>       * This indicates to me that the depth/complexity of the crafted resource in the test is too high.
> 
>     * However, I do think it is a good benchmark for the autocrafting code.
> 
>     * Right now it isolates crafting complexity. To be even more realistic, it would need to have a bigger root storage to also take the copying cost into account.

That's correct so far. I'm not sure how well that is covered right now, but having the same ingredient multiple times (and the same for patterns for that ingredient) blows up the search when resources are missing, because (more or less) the same search is done multiple times.

raoulvdberge:
Regarding this message https://github.com/refinedmods/refinedstorage2/pull/1129#issuecomment-3220518111 I'm not sure I can follow this statement:

> This seems odd, not sure if this is the intended behavior. Especially as we had mixed planks before already! I think MISSING_RESOURCES should only be the result if all possibilities together cannot satisfy the needs. But I wonder if crafting should be preferred over using multiple different resources in that case. If not, we would need to track craftable items and try them after all available resources were checked/consumed.

MISSING_RESOURCES is only returned if none of the ingredients work, currently.

Yes, if you combine ingredients then it will no longer differentiate between slots and MISSING_RESOURCES could be a misleading result.

Crafting should not be preferred over using multiple resources. But how would tracking craftable items solve this issue, if you end up combining all the inputs?

> a possible change would need to somehow keep track of partial results (i.e., first try to satisfy the requirement using multiple resources if multiple variants are allowed, not just one).

I don't understand this. Currently, it's checking multiple resources already.

> I'm not sure how well that would work with fuzzy processing patterns, when e.g., 2 stacks of logs are required: With this change, we would allow stuff like 1x oak log + 127 spruce log. Would it be a requirement to have 2 stacks of 1 material each in this case?

It's a good question. My assumption is that mixing within 1 slot will give trouble.

raoulvdberge:
I'm struggling to wrap my head around https://github.com/refinedmods/refinedstorage2/pull/1129/commits/d056b19f50fa80ebb70e66641bad26b30fe1dd8d

Won't calculating the resources in storage before give problems?
Let's say we require an ingredient with possibilities [A,B].

We first calculate A, we do a deep subcraft and we mutate the internal storage.
We run out of items for A, so we cycle the ingredient and try B.

In the old code, this would calculate the resources in storage **at this point**.

However, in the new code, the amounts are pre-calculated. 

But wouldn't this sorting be incorrect, not accounting of the changes being made for A?

Is there something that I'm missing?

SirYwell:
> Regarding this message [#1129 (comment)](https://github.com/refinedmods/refinedstorage2/pull/1129#issuecomment-3220518111) I'm not sure I can follow this statement:

Sorry for the confusion, this is indeed about the case when multiple ingredients that allow the same resources are combined.

> Yes, if you combine ingredients then it will no longer differentiate between slots and MISSING_RESOURCES could be a misleading result.

> It's a good question. My assumption is that mixing within 1 slot will give trouble.

I think that's the underlying problem. But it *might* be safe for crafting patterns. In that case, fixing the calculation based on my deduplication changes would need two code paths, one for crafting (="it is fine to use different materials for this ingredient") vs processing (="one slot must only use one material"). That might get quite complex.

> Crafting should not be preferred over using multiple resources. But how would tracking craftable items solve this issue, if you end up combining all the inputs?

I'd need to investigate, I might be wrong there. But if I want to craft 3x crafting table, even currently, we look for 3x planks for each slot. And for each slot, only 1 material is allowed (iirc). In that case, we craft planks instead of trying to use planks of different types for that slot for the different crafting tables. With ingredient deduplication, due to the higher amount needed, we also try to craft more (and if we can't, we fail). **If** using multiple materials for one (deduplicated) ingredient is allowed, it would make sense to first use the available materials (and mix them). And if that's not enough, try crafting more of any of them. "Tracking" in some sense would be a heuristic for which material to start with. I'm not sure if that makes any sense though.

I hope that makes it a bit clearer. There are many difficult details involved, making it really hard to explain, sorry.

SirYwell:
> I'm struggling to wrap my head around [d056b19](https://github.com/refinedmods/refinedstorage2/commit/d056b19f50fa80ebb70e66641bad26b30fe1dd8d)
> 
> Won't calculating the resources in storage before give problems? Let's say we require an ingredient with possibilities [A,B].
> 
> We first calculate A, we do a deep subcraft and we mutate the internal storage. We run out of items for A, so we cycle the ingredient and try B.
> 
> In the old code, this would calculate the resources in storage **at this point**.
> 
> However, in the new code, the amounts are pre-calculated.
> 
> But wouldn't this sorting be incorrect, not accounting of the changes being made for A?
> 
> Is there something that I'm missing?

Hm, the sorting shouldn't be affected:
It happens at the same time at the same place. The only difference is that we look up the amount directly before sorting, instead of during sorting. But in between, there is nothing modifying the amount, right?

However, regarding cycling: If the internal storage changes the amount of B due to A, then B would give an outdated result. I'm not sure if that can happen? Similarly, if A runs into crafting something that requires B, then we'd modify the storage to have less of B. Could that happen?

But if such effects should be reflected in the array order, then you would need to re-sort the tail of the array every time I think?

raoulvdberge:
I was trying to write a test to prove that it could break, but I couldn't. Indeed, it can never happen, for this reason:

> But wouldn't this sorting be incorrect, not accounting of the changes being made for A?

We don't care about the changes for A, as the subcraft for A has failed (there were not enough resources to craft A).
So the intermediates for A are dismissed and it will try B.
The system does not handle partial crafts (doing that as well would only make things slower), the full subcraft has to succeed.
There is nothing that A could do that could affect the sorting or inStorage counts later.

Given that, the sorting can perfectly be predetermined.

And given that, I'll merge this PR. Thanks a lot!

PR #1134, opened 2025-08-26
Give storage disks a new id when duplicating it in creative mode

### LINKED ISSUE #918: Look for a way to avoid duplicate disks in drives
### Describe your enhancement

ideas:
- prevent dupe disks in the same drive?
- prevent dupe disks in the same network?
- prevent dupe disks in the same world?
- prevent duping disks via creative copy

starforcraft:
Partially fixes https://github.com/refinedmods/refinedstorage2/issues/918

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+-   Duplicating storage disks with middle click in creative mode will now give a new storage disk instead of referencing the copied one.
+
 ## [2.0.0-beta.11] - 2025-08-25
 
 ### Added
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/storage/StorageContainerItemHelper.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/storage/StorageContainerItemHelper.java
@@ -49,4 +49,6 @@ public interface StorageContainerItemHelper {
     void transferFromBlockEntity(ItemStack stack, StorageBlockEntity blockEntity);
 
     void markAsToTransfer(ItemStack from, ItemStack to);
+
+    boolean clear(ItemStack stack);
 }
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/StorageContainerItemHelperImpl.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/StorageContainerItemHelperImpl.java
@@ -199,6 +199,16 @@ public class StorageContainerItemHelperImpl implements StorageContainerItemHelpe
         return Optional.ofNullable(stack.get(DataComponents.INSTANCE.getStorageReference()));
     }
 
+    @Override
+    public boolean clear(final ItemStack stack) {
+        if (getId(stack).isEmpty()) {
+            return false;
+        }
+
+        stack.remove(DataComponents.INSTANCE.getStorageReference());
+        return true;
+    }
+
     private Optional<UUID> getIdToBeTransferred(final ItemStack stack) {
         return Optional.ofNullable(stack.get(DataComponents.INSTANCE.getStorageReferenceToBeTransferred()));
     }
--- a/refinedstorage-fabric/src/main/java/com/refinedmods/refinedstorage/fabric/mixin/AbstractContainerMenuMixin.java
+++ b/refinedstorage-fabric/src/main/java/com/refinedmods/refinedstorage/fabric/mixin/AbstractContainerMenuMixin.java
@@ -1,0 +1,63 @@
+package com.refinedmods.refinedstorage.fabric.mixin;
+
+import com.refinedmods.refinedstorage.common.api.RefinedStorageApi;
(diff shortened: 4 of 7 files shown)


raoulvdberge:
Why partially? Ok there's no check on network level but I think this is pretty good already

starforcraft:
I wrote 'partially fixes' because it's still possible to have duplicated storage disk uuids via other methods. I think it would help to add some kind of warning when the disk drive contains disks with the same UUID. This would improve accessibility and also prevent players from submitting reports about items magically multiplying.
Also just not reading the same UUID more than once would be the best case scenario

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/storage/StorageContainerItemHelper.java
```diff
@@ -49,4 +50,8 @@ void appendToTooltip(ItemStack stack,
     void transferFromBlockEntity(ItemStack stack, StorageBlockEntity blockEntity);
 
     void markAsToTransfer(ItemStack from, ItemStack to);
+
+    Optional<UUID> getId(ItemStack stack);
+
+    void removeId(ItemStack stack);
```
raoulvdberge: IDs are supposed to be an implementation detail.
Rather expose something like `clear` (that returns a bool if something was cleared, for use in the mixin).

refinedstorage-fabric/src/main/java/com/refinedmods/refinedstorage/fabric/mixin/AbstractContainerMenuMixin.java
```diff
@@ -0,0 +1,60 @@
...
+
+        final ItemStack stack = slot.getItem();
+        final StorageContainerItemHelper helper = RefinedStorageApi.INSTANCE.getStorageContainerItemHelper();
+
+        if (helper.getId(stack).isEmpty()) {
+            return;
+        }
+
+        helper.removeId(stack);
+        this.setCarried(stack.copyWithCount(stack.getMaxStackSize()));
```
raoulvdberge: Isn't copy before removing safer?
```suggestion
        final ItemStack copy = stack.copy();
        if (helper.clear(copy)) {
            this.setCarried(copy.copyWithCount(stack.getMaxStackSize());
            return;
        }
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+-   Duplicating storage disks in creative mode is now possible because each copy is assigned a new UUID
```
raoulvdberge:
```suggestion
-   Duplicating storage disks with middle click in creative mode will now give a new storage disk instead of referencing the copied one.
```

raoulvdberge (changes requested):

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/storage/StorageContainerItemHelperImpl.java
```diff
@@ -195,10 +195,20 @@ public void markAsToTransfer(final ItemStack from, final ItemStack to) {
         getId(from).ifPresent(id -> to.set(DataComponents.INSTANCE.getStorageReferenceToBeTransferred(), id));
     }
 
-    private Optional<UUID> getId(final ItemStack stack) {
+    public Optional<UUID> getId(final ItemStack stack) {
```
raoulvdberge: Now make it private again

refinedstorage-fabric/src/main/java/com/refinedmods/refinedstorage/fabric/mixin/AbstractContainerMenuMixin.java
```diff
@@ -0,0 +1,60 @@
...
+
+        final Slot slot = this.slots.get(slotId);
+        if (!slot.hasItem()) {
+            return;
+        }
+
+        final ItemStack stack = slot.getItem();
+        final StorageContainerItemHelper helper = RefinedStorageApi.INSTANCE.getStorageContainerItemHelper();
+
+        if (helper.getId(stack).isEmpty()) {
```
raoulvdberge: I guess the mixin needs to be updated here

refinedstorage-neoforge/src/main/java/com/refinedmods/refinedstorage/neoforge/mixin/AbstractContainerMenuMixin.java
```diff
@@ -0,0 +1,58 @@
...
+        if (!slot.hasItem()) {
+            return;
+        }
+
+        final ItemStack stack = slot.getItem();
+        final StorageContainerItemHelper helper = RefinedStorageApi.INSTANCE.getStorageContainerItemHelper();
+
+        final ItemStack copy = stack.copy();
+        if (helper.clear(copy)) {
+            this.setCarried(copy.copyWithCount(stack.getMaxStackSize()));
```
raoulvdberge, line 59: Since this is quite an invasive mixin, perhaps it's a good idea to log here if we cleared something.

raoulvdberge (changes requested):
Fabric build is broken

raoulvdberge (approved):

PR #1135, opened 2025-08-29
perf: prevent unnecessary redstone checks in network node logic

MerelSollie:
### **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
This PR introduces a performance improvement in the logic of the redstone activeness check of ``com.refinedmods.refinedstorage.common.support.network.AbstractBaseNetworkNodeContainerBlockEntity@calculateActive()`` .

### **What is the current behavior?** (You can also link to an open issue here)
In the current logic, the Base BlockEntity for all network node blocks, (including but not limited to cables, controllers, importers, etc.) will check and request if the block is being powered by redstone, even if that block has a Redstone Active mode of IGNORE.

This means that blocks like cables or importers which have the mode IGNORE, and which might not even be changed from IGNORE to HIGH/LOW, will perform many costly calculations to check if what the value of the redstone signal is, despite the results of this check not being used later in the process, as IGNORE is hardcoded to return ``true``.

This can mean that even on a simple flat world with a lot of cables and no further blocks, a large performance overhead is present due to the many redstone activeness checks.

### **What is the new behavior?**
In the new logic, the hardcoded check for redstone active mode IGNORE is placed in ``AbstractBaseNetworkNodeContainerBlockEntity@calculateActive`` before the code to request the current redstone signal is executed.

In practice this means the performance overhead of cables and other blocks, like importers, can be significantly decreased. In the playtesting on our world this resulted in a decrease in the time spend per tick, which saved around 4 seconds worth of checks to determine the redstone status of these blocks per minute. Which is around 1.68 processing hours per 24h day saved in unnecessary checks.

### **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
This PR does not introduce a breaking change, as it only slightly changes the order in which the behaviour is determined. It does not impact any features or behaviour.

### **Other information**:
There is no open issue as far as I am aware. I had introduced  the fix already, so I figured I'd save the time writing an issue.

We discovered this performance issue after we began running into issues of server tick rate after most our friend group had migrated to refined storage to access Sophisticated Storage inventories over a longer distance with a combination of crafting grids and external interfaces.

Upon tests using profilers and multiple test setups, including unpowered networks, we found the issue to be with the ``level.hasNeighborSignal(worldPosition));`` call. We were able to replicate this in a clean install of the mod and test worlds, and found the line that is the culprit.

The fix itself is not the cleanest, as it does technically leave the check in  ``com.refinedmods.refinedstorage.common.support.RedstoneMode@isActive()`` as well. However, I've left the old behaviour in there as well so ``com.refinedmods.refinedstorage.common.storage.portablegrid.InWorldPortableGrid@isGridActive()`` wouldn't need to be changed. If you want me to change anything, or move it into a method in the RedstoneMode enum, I'd be happy to do that! :slightly_smiling_face: 

I've also not added this to the release notes, as I am not sure whether this technical change would be relevant enough for the end-user.

### DIFF
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/detector/DetectorBlockEntity.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/detector/DetectorBlockEntity.java
@@ -181,4 +181,10 @@ public class DetectorBlockEntity extends AbstractBaseNetworkNodeContainerBlockEn
                                                                    final BlockState newBlockState) {
         return AbstractDirectionalBlock.didDirectionChange(oldBlockState, newBlockState);
     }
+
+    @Override
+    protected boolean hasRedstoneMode() {
+        return false;
+    }
+
 }
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/networking/AbstractCableBlockEntity.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/networking/AbstractCableBlockEntity.java
@@ -20,4 +20,9 @@ public abstract class AbstractCableBlockEntity extends AbstractCableLikeBlockEnt
     public Component getName() {
         return overrideName(ContentNames.CABLE);
     }
+
+    @Override
+    protected boolean hasRedstoneMode() {
+        return false;
+    }
 }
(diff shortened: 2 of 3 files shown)


raoulvdberge:
Good stuff. Wouldn't it be a good idea to add `boolean hasRedstoneMode() { return false; }` to `AbstractCableBlockEntity` and `DetectorBlockEntity`?

MerelSollie:
> Good stuff. Wouldn't it be a good idea to add `boolean hasRedstoneMode() { return false; }` to `AbstractCableBlockEntity` and `DetectorBlockEntity`?

Hi! Thanks for checking this PR! 

I could override that method as well, no problem. But despite that I'd think it would be wise to also keep the existing change already present in this PR. 

When coming up with a solution I thought about that option. Whilst overriding the ``hasRedstoneMode()`` to return ``false`` would be a bit cleaner for those blocks. But that way a ton of processing would still be wasted on blocks like grids, importers, controllers etc. if they have the redstoneMode set to ignore. 

However it would add a backstop for cables and such at least, so I'll add that as well.

MerelSollie:
I added the ``hasRedstoneMode`` overrides. This mainly affects the time and data spent storing the redstone modes for these block types. I didn't observe any additional performance improvements beyond that.

raoulvdberge:
Well yes, my suggestion was meant to be in combination with your existing change.

MerelSollie:
> Well yes, my suggestion was meant to be in combination with your existing change.

Alright, great! In that case, that's what I added in the last commit :slightly_smiling_face:

raoulvdberge:
Merged, thanks.

Issue #987, opened 2025-04-29
TPS-Expensive auto-crafting


naxane:
### Describe the bug

Ran some profiling while looking at another issue. Auto-crafting calculations appear TPS-expensive.

### How can we reproduce this bug or crash?

[YUvzoVbPMl.sparkprofile.txt](https://github.com/user-attachments/files/19958221/YUvzoVbPMl.sparkprofile.txt)

### What Minecraft version is this happening on?

Minecraft 1.21.1

### What NeoForge or Fabric version is this happening on?

NeoForge 21.1.148

### What Refined Storage version is this happening on?

2.0.0-beta.2

### Relevant log output

```shell

```

raoulvdberge:
<img width="1222" height="528" alt="Image" src="https://github.com/user-attachments/assets/291844f6-c288-4d80-8550-96b92dcd3b1c" />

Primary causes seems to be constantly checking the max amount that we can craft.
We have to check how much we can craft, as something like the interface could request more of a resource than is currently available.
(The interface could require 64 of something, but we can only craft 8. If we don't correct it to only craft 8, the entire calculation would fail due to missing resources and nothing would happen.)

Even if we get rid of that max amount calculation somehow, constantly calculating tasks will be slow too. 

There are a few fixes that could be done:

- We could use some cooldown if missing resources constantly happen (see for example `ExternalStorageWorkRate`) 
  - This is pretty difficult to do as we are already throttled within the exporter work rate itself. 
- We should only calculate the max amount if a request for the initial amount fails.
- We should only calculate the max amount until the amount we requested initially (binary search low until initial requested amount), and not keep going higher.
  - This is not necessary anymore if we only calculate the max amount if a request for the initial amount fails.
  - Why? For most cases, there will no longer be an unbounded max amount check, but rather a max amount check that will go up to N. Where N is the typical transfer quota of an exporter or interface (around 64 items).
  - It will not be unbounded because we know for sure that there are missing resources.

raoulvdberge:
I did a fix for this for beta 12. Let me know how it works.

SirYwell:
One observation: The autocrafting currently still walks all the crafting trees, even if one already failed. From my understanding, it should be possible to exit early if a subtree cannot be crafted, if we don't need a list of missing resources.

I've built an extreme example of a Mega Exporter from Cable Tiers trying to export 1x1048m Storage Part, with mostly missing resources but with crafting patterns for all previous storage part tiers.
[...]

PR #1142, opened 2025-09-07
perf: skip missing resource calculation when unneeded

SirYwell:
As written in #987, we can take a shortcut if we don't need a complete list of missing resources.

### DIFF
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingCalculatorListener.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/calculation/CraftingCalculatorListener.java
@@ -25,4 +25,11 @@ public interface CraftingCalculatorListener<T> {
     }
 
     T getData();
+
+    /**
+     * {@return true if missing resource calculation must be complete}
+     */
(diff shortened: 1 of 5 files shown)


raoulvdberge:
Mutation testing is failing

SirYwell:
Hm, I think that's because it's failing on `develop` already?

raoulvdberge:
Ha, yes, looks like I caused it. Let me check.

raoulvdberge:
I've rebased and force pushed this branch, but seems to still be failing.
On develop it's succeeding now.

SirYwell:
I'll take a closer look tomorrow.

SirYwell:
I don't see a good way to fix the mutation testing here. On `develop`, we are at 90% (411/459) mutation coverage, with my changes we are at 89% (414/463) mutation coverage. There are 2 mutations that aren't killed: returning true instead of false in TaskPlanCraftingCalculatorListener and IsCraftableCraftingCalculatorListener `requiresFullMissingResourcesCalculation`. But these mutations don't affect correctness, only performance. The API also doesn't really expose the actual listeners, only their result, so we can't even test for that.

raoulvdberge:
Yeah, this is a typical case where mutation testing doesn't really work well (behavior is the same but performance is different). Let me take a look.

PR #1207, opened 2025-10-27
perf: lazily compute PreviewBuilder state

SirYwell:
Trying to re-use TreePreviewCraftingCalculatorListener to calculate the normal Preview too, as that seems to have better performance characteristics.

### DIFF
--- a/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/preview/PreviewBuilder.java
+++ b/refinedstorage-autocrafting-api/src/main/java/com/refinedmods/refinedstorage/api/autocrafting/preview/PreviewBuilder.java
@@ -5,22 +5,30 @@ import com.refinedmods.refinedstorage.api.core.CoreValidations;
 import com.refinedmods.refinedstorage.api.resource.ResourceAmount;
 import com.refinedmods.refinedstorage.api.resource.ResourceKey;
 
+import java.util.ArrayDeque;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.jspecify.annotations.Nullable;
+
 public class PreviewBuilder {
     private final Map<ResourceKey, MutablePreviewItem> items;
     private List<ResourceAmount> outputsOfPatternWithCycle = Collections.emptyList();
     private boolean missing;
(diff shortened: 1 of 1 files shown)


SirYwell:
@raoulvdberge what do you think of this in general? Currently the order is different, I didn't take a closer look if that can be fixed easily. But it seems to be faster and would render the PreviewCraftingCalculatorListener and related code unnecessary.

raoulvdberge:
No, I'm not a fan of this. The code should match the intent, and now you are reusing tree code for list code because it is faster. I'd like to know WHY it's faster, and use that knowledge to improve the list code.

I have no problem with having both list and tree code.

For me the ordering in the list is important.

I don't want the list view to become dependant on the tree view, the whole listener system is set in place to avoid that. 
If lists end up getting different requirements in the future, we'll be in a world of hurt to retrofit the tree code to support it too.

SirYwell:
Understandable.
The main bottleneck from the `PreviewBuilder` is the copy, because we copy a lot of data many times with just little changes.
This isn't a problem with `MutableTreePreviewNode`. That's why it is faster when reducing the `MutableTreePreviewNode` at the end to a `Preview`.

Maybe the concept can be used for PreviewBuilder too, e.g., remember the relevant `PreviewBuilder` instances and instead of copying their data each time, merge them at the end. But that also might have difficulties when we want to maintain the exact same order, not sure. Other alternatives would include using immutable collections, but they typically have asymptotically slower access and the value would need to be immutable too, so not sure if there's any benefit from that (and it would require an additional dependency).

SirYwell:
I now reworked the implementation of `PreviewBuilder` directly to avoid the copy and remember a trace of `PreviewBuilder`s. This allows us to do all the necessary computation in one step at the end (but in theory at any moment, for that current state). The correct order *should* be maintained now too if I didn't miss something.

I didn't test this beyond existing unit tests for now. I also didn't profile it.

raoulvdberge:
@SirYwell  Can we get this PR in?

SirYwell:
> @SirYwell Can we get this PR in?

Should be ready now from my side.

raoulvdberge:
Please rebase on develop. A changelog entry is not necessary here.

raoulvdberge:
Github was still complaining your branch is not up to date, so I updated it through the UI.

raoulvdberge:
Thanks! <3

PR #1242, opened 2026-01-30
fix: containermenu slot lists do not get cleared

### LINKED ISSUE #1240: Slot lists do not get cleared correctly
### Describe the bug

In MC ContainerMenus slots are stored in 3 lists
public final NonNullList<Slot> slots = NonNullList.create();
private final NonNullList<ItemStack> remoteSlots;
private final NonNullList<ItemStack> lastSlots = NonNullList.create();

When slots are reloaded during a resize, slots is cleared, but remoteSlots and lastSlots are not and thus accumulate members.

Add a breakpoint inside resize() and look at lastSlots and remoteSlots. when resizing the grid the member number increases.

Minecraft 1.21.1

all of them

2.0.0

### What modpack and modpack version, if any, are you using?

_No response_

### Relevant log output

```shell

```

lucasser:
fixes #1240 

A similar fix might be needed for dataSlots and remoteDataSlots but I didn't look into those.


raoulvdberge:
I've implemented a fix without mixins in da623bda

Issue #1259, opened 2026-03-05
Resources dont get filtered by display name in grid


lucasser:
### Describe the bug

When searching in grid for an item the filter operates on the item's name such as minecraft:paper. And if the item is renamed, the display name is not queried which is counterintuitive.

### How can we reproduce this bug or crash?

rename item
search for it

### What Minecraft version is this happening on?

Minecraft 1.21.1

### What NeoForge or Fabric version is this happening on?

doesn't matter

### What Refined Storage version is this happening on?

2.0.1

### Relevant log output

```shell

```

raoulvdberge:
This was fixed in Refined Storage v3.0.3 (for Minecraft 26.1.2) and v2.0.7 (for Minecraft 1.21.1). Thank you.

PR #1260, opened 2026-03-05
fix: searching checks both item name and display name

lucasser:
Added an itemName and a name variable to AbstractGridResource

If there is a stack of paper named "dinnerbone"
To get the item name "Paper" use resource.getItemName()
To get the display name "dinnerbone" use resource.getDisplayName()

GridQueryParser.parseLiteral() checks both itemName and displayName.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+-   Searching in the Grid now checks against the resource name, registry id, and display name (for renamed items)
+
 ## [3.0.0-beta.4] - 2026-04-16
 
 ### Added
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
@@ -5,6 +5,7 @@ import com.refinedmods.refinedstorage.api.resource.repository.ResourceRepository
 import com.refinedmods.refinedstorage.api.storage.tracked.TrackedResource;
 import com.refinedmods.refinedstorage.common.api.support.resource.PlatformResourceKey;
 
+import java.util.List;
 import java.util.Set;
 import java.util.function.Function;
 
@@ -14,14 +15,14 @@ import org.jspecify.annotations.Nullable;
 @API(status = API.Status.STABLE, since = "2.0.0-milestone.3.0")
 public abstract class AbstractGridResource<T extends PlatformResourceKey> implements GridResource {
     protected final T resource;
-    private final String name;
+    private final List<String> names;
     private final Function<GridResourceAttributeKey, Set<String>> attributes;
 
     protected AbstractGridResource(final T resource,
-                                   final String name,
+                                   final List<String> names,
                                    final Function<GridResourceAttributeKey, Set<String>> attributes) {
         this.resource = resource;
-        this.name = name;
+        this.names = names;
         this.attributes = attributes;
     }
 
@@ -39,8 +40,13 @@ public abstract class AbstractGridResource<T extends PlatformResourceKey> implem
     }
 
     @Override
-    public String getName() {
-        return name;
+    public List<String> getSearchableNames() {
+        return names;
+    }
+
+    @Override
+    public String getHoverName() {
+        return names.getLast();
     }
 
     @Override
@@ -63,7 +69,7 @@ public abstract class AbstractGridResource<T extends PlatformResourceKey> implem
     public String toString() {
         return "AbstractGridResource{"
             + "resource=" + resource
-            + ", name='" + name + '\''
+            + ", valid names='" + getSearchableNames() + '\''
             + ", attributes=" + attributes
             + '}';
     }
--- a/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/GridResource.java
+++ b/refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/GridResource.java
@@ -31,7 +31,9 @@ public interface GridResource {
 
     long getAmount(ResourceRepository<GridResource> repository);
 
-    String getName();
+    List<String> getSearchableNames();
+
+    String getHoverName();
 
     Set<String> getAttribute(GridResourceAttributeKey key);
 
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/AbstractItemGridResourceRepositoryMapper.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/AbstractItemGridResourceRepositoryMapper.java
@@ -34,6 +34,7 @@ public abstract class AbstractItemGridResourceRepositoryMapper implements Resour
         final ItemResource itemResource = (ItemResource) resource;
         final Item item = itemResource.item();
         final ItemStack itemStack = itemResource.toItemStack();
+        final String hoverName = itemStack.getHoverName().getString();
         final String name = item.getName(itemStack).getString();
         final String modId = getModId(itemStack);
         final String modName = getModName(modId).orElse("");
@@ -46,6 +47,7 @@ public abstract class AbstractItemGridResourceRepositoryMapper implements Resour
         return new ItemGridResource(
             itemResource,
             itemStack,
+            hoverName,
             name,
             k -> attributes.getOrDefault(k, Collections::emptySet).get()
         );
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/FluidGridResource.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/FluidGridResource.java
@@ -41,7 +41,9 @@ public class FluidGridResource extends AbstractGridResource<FluidResource> {
     public FluidGridResource(final FluidResource resource,
                              final String name,
                              final Function<GridResourceAttributeKey, Set<String>> attributes) {
-        super(resource, name, attributes);
+        super(resource,
+            List.of(BuiltInRegistries.FLUID.getKey(resource.fluid()).getPath(), name),
+            attributes);
         this.id = BuiltInRegistries.FLUID.getId(resource.fluid());
         this.rendering = RefinedStorageClientApi.INSTANCE.getResourceRendering(FluidResource.class);
     }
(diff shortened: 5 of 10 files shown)


raoulvdberge:
This would fix #1259, right?

raoulvdberge:
I'm not really fan of this PR, it introduces 3 variants of "names" in the code. And I have no idea how this fixes the problem of RS searching on the registry name "minecraft:paper", like mentioned in the original issue? Is that something that is fixed here? I see `getName` on `Item` is still being used.

lucasser:
Yes, it solves #1259 
This will keep track of two different names, the item registry id/base item name, and the display name.
In most cases the two are the same, unless the item gets renamed.
The parser
```java
        return (repository, resource) -> normalize(resource.getItemName()).contains(normalize(node.token().content()))
            || normalize(resource.getDisplayName()).contains(normalize(node.token().content()));
```

looks at both names. So if you have a renamed piece of paper it'll come up when you search paper (The current behaviour, useful if you are sorting through renamed Apotheosis gear for example) but also when you search for its name

raoulvdberge:
Question, in your PR description you mention:
> To get the item name "Paper" use resource.getItemName()
> To get the display name "dinnerbone" use resource.getDisplayName()

But in the message above, you talk about the registry name. That is not mentioned in your PR description, it's a bit confusing.

I would probably change the logic a bit:

- Return a `Set<String>` instead of adding more "getName"-esque methods (these "getName"-esque methods are not always applicable to all resource types)
- Search on the item name, display name, and registry ID.

raoulvdberge:
Could you rebase on the latest `develop` please?

raoulvdberge:
Also, please create these lists in the constructor. We don't want a bunch of list allocations each time you search something..

lucasser:
Done

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
```diff
@@ -63,7 +69,7 @@ public PlatformResourceKey getResourceForRecipeMods() {
     public String toString() {
         return "AbstractGridResource{"
             + "resource=" + resource
-            + ", name='" + name + '\''
+            + ", name='" + getName() + '\''
```
raoulvdberge: Just dump the names list here (regenerate `toString`)

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/GridResource.java
```diff
@@ -31,6 +31,8 @@ public interface GridResource {
 
     long getAmount(ResourceRepository<GridResource> repository);
 
+    List<String> getNames();
+
     String getName();
```
raoulvdberge: Why do we still have this one? Maybe give it a better name.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/AbstractItemGridResourceRepositoryMapper.java
```diff
@@ -34,7 +34,8 @@ public GridResource apply(final ResourceKey resource) {
         final ItemResource itemResource = (ItemResource) resource;
         final Item item = itemResource.item();
         final ItemStack itemStack = itemResource.toItemStack();
-        final String name = item.getName(itemStack).getString();
+        final String displayName = itemStack.getHoverName().getString();
```
raoulvdberge: Name it `hoverName`.

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/FluidGridResource.java
```diff
@@ -39,9 +39,11 @@ public class FluidGridResource extends AbstractGridResource<FluidResource> {
     private final ResourceRendering rendering;
 
     public FluidGridResource(final FluidResource resource,
-                             final String name,
+                             final String fluidName,
```
raoulvdberge: Yes, it's a `FluidGridResource` no need to put `fluid` in this var name again

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParserImplTest.java
```diff
@@ -78,6 +79,17 @@ void testNameQuery(final String query) throws GridQueryParserException {
         assertThat(predicate.test(repository, new R("Glass"))).isFalse();
     }
 
+    @ParameterizedTest
+    @ValueSource(strings = {"dirt", "Dirt", "DiRt", "Di", "irt", "test", "TeSt"})
+    void testNameQueryRenamedItems(final String query) throws GridQueryParserException {
```
raoulvdberge, line 83: How does this work, the test setup hasn't changed? What does this test add

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParserImplTest.java
```diff
@@ -288,27 +300,31 @@ void testInvalidTokenInUnaryCountQuery(final String operator) {
 
     private static class R implements GridResource {
         private final String name;
+        private final String resourceName;
         private final long amount;
         private final Map<GridResourceAttributeKey, Set<String>> attributes;
 
-        R(final String name) {
-            this(name, 1);
+        R(final String name, final String resourceName) {
```
raoulvdberge: name and resourceName ?

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+-   Searching in grid checks against the item name and the display name
```
raoulvdberge: Yes and also the id right?

raoulvdberge (changes requested):

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParserImplTest.java
lucasser: It tests for the renamed search. `assertThat(predicate.test(repository, new R(List.of("Dirt", "test")))).isTrue();` has been renamed to test.

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/GridResource.java
lucasser: I renamed them to getSearchableNames and getHoverName. getHoverName is only used in tests to check if the items are correct. Should i just remove it then and have the tests check getSearchableNames().last instead?

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParserImplTest.java
lucasser: As in the item used in that assert has been renamed to test, and should show up for all the search strings

refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/view/AbstractItemGridResourceRepositoryMapper.java
```diff
@@ -34,7 +34,8 @@ public GridResource apply(final ResourceKey resource) {
         final ItemResource itemResource = (ItemResource) resource;
         final Item item = itemResource.item();
         final ItemStack itemStack = itemResource.toItemStack();
-        final String name = item.getName(itemStack).getString();
+        final String hoverName = itemStack.getHoverName().getString();
+        final String itemName = item.getName(itemStack).getString();
```
raoulvdberge, line 38:
```suggestion
        final String name = item.getName(itemStack).getString();
```

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParserImplTest.java
raoulvdberge: Just consolidate this in the other query test.

refinedstorage-common/src/test/java/com/refinedmods/refinedstorage/common/grid/GridSortingTypesTest.java
```diff
@@ -85,28 +85,28 @@ void testSortingAscending(final GridSortingTypes sortingType) {
...
-                .extracting(GridResource::getName)
+                .extracting(GridResource::getHoverName)
                 .containsExactly(
                     "Dirt",
                     "Gold Ingot",
                     "Stone"
                 );
             case ID -> assertThat(repository.getViewList())
-                .extracting(GridResource::getName)
+                .extracting(resource -> resource.getSearchableNames().getLast())
```
raoulvdberge, line 102: Why getLast here? instead of hoverName

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+-   Searching in grid checks against the resource name, registry id, and display name (for renamed items)
```
raoulvdberge, line 12:
```suggestion
-   Searching in the Grid now checks against the resource name, registry id, and display name (for renamed items)
```

raoulvdberge (changes requested):

lucasser:
From one of the review comments:
> I renamed them to getSearchableNames and getHoverName. getHoverName is only used in tests to check if the items are correct. Should i just remove it then and have the tests check getSearchableNames().last instead?

Do i keep the method just for tests, or get rid of it?

refinedstorage-common-api/src/main/java/com/refinedmods/refinedstorage/common/api/grid/view/AbstractGridResource.java
```diff
@@ -63,7 +69,7 @@ public PlatformResourceKey getResourceForRecipeMods() {
     public String toString() {
         return "AbstractGridResource{"
             + "resource=" + resource
-            + ", name='" + name + '\''
+            + ", valid names='" + getSearchableNames() + '\''
```
raoulvdberge, line 72:
```suggestion
            + ", names='" + names + '\''
```

raoulvdberge (changes requested):

raoulvdberge:
> Do i keep the method just for tests, or get rid of it?

I'd get rid of it

raoulvdberge:
Can you make it so that `AbstractGridResource` stays compatible for addons? (add a constructor for `String`)

raoulvdberge:
I've fixed this myself in 10d84098, with some naming changes and backwards compat for addons (compatibility constructor).

PR #1318, opened 2026-05-03
Feature: Network Controller display Network Nodes and their energy consumption

OskarPetersilie:
Hey since v2 doesnt include the energy consumption display i forked it and added the feature.

If it fits feel free to merge.

I tryd to make it as less perfomance hungry as possible and stick to the patterns the projects used. I am by far not a java developer so it may be also shitty code. 

I checked that is displays correct values.

image

Also added fitting tooltips

image

image


raoulvdberge:
Consult me before introducing new features.

The ability to add multiple controllers to a network, effectively increasing total energy storage, changes the role of the controller. It is no longer a centralized network block. The device list was removed intentionally to reflect that the controller now serves only as an energy component. Reintroducing the list is planned under #81.

PR #1402, opened 2026-09-11
perf: avoid repeated query normalization in Grid search

### LINKED ISSUE #1401: Grid search is slower than it needs to be on large networks
### Describe the bug

With 37,377 different resources on one disk, every character typed in the Grid search box costs (on my machine):

- `pickaxe`: 4.5 ms
- `@minecraft`: 15.8 ms
- `#tool`: 21.6 ms

I measured this by adding a timer around `AbstractGridContainerMenu.onSearchTextChanged`, which covers the parse plus `setFilterAndSort`. It runs 10 times per key press and I take the fastest one, so a GC or a render hitch does not skew it. Typing was scripted at one character per second.

This is done on the Render thread, so this can cost fps.

I looked into it and found `GridQueryParser` to be part of the cause. The search text is lowercased again for every resource, even though it never changes during one search (`parseLiteral` line 182, `attributeMatch` line 169). `attributeMatch` also builds a new stream per resource.

Lowercasing the search text once, before the loop, and replacing the stream with a plain loop gives the "after" column below. Row counts stay identical on every key press.

There may well be more to it, this is only the part I could pin down. I have the change ready and am happy to open a PR to show my proposal, but needed an issue first.

1. Disk Drive, Grid, Creative Storage Disk.
2. Fill the disk with the debug stick while crouching. About 37,000 resources.
3. Open the Grid, type `#tool` one character at a time.
4. Every key press spends over 20 ms in `setFilterAndSort` on the Render thread (on my machine).

Minecraft 26.1.x

NeoForge 26.1.2.78

Development build of `develop` at 0643e6f4b

### What modpack and modpack version, if any, are you using?

None

### Relevant log output

```shell
query          rows     before    after
p              12543      8.59     7.88
pi              4322      5.92     3.99
pic             4261      4.31     3.94
pick            4255      4.34     3.94
picka           4254      4.27     4.03
pickax          4254      4.27     3.53
pickaxe         4254      5.00     3.55
(cleared)      37377      7.75     8.12
@m             36972     18.59    10.73
@mi            36972     17.88    11.42
@min           36972     15.77    13.23
@mine          36972     16.07    10.85
@minec         36972     15.52    10.86
@minecr        36972     16.05    10.80
@minecra       36972     16.00    10.68
@minecraf      36972     15.63    10.74
@minecraft     36972     15.65    10.73
(cleared)      37377      7.83     8.14
#t             36647     20.46    16.24
#to            28290     23.30    14.60
#too           28018     20.91    14.40
#tool          28018     20.71    14.60
(cleared)      37377      7.71     8.18
```

The cleared lines show the cost of the sort alone, about 7.8 ms, so the filtering is the part above that. Those lines are slightly slower in the after run, which is just noise on my machine, and it runs against the change rather than for it.

Klaas058:
Closes #1401

`GridQueryParser` normalized the search text inside the per resource lambda, and `attributeMatch` built a stream per resource, so a constant string was trimmed and lowercased once per resource per keystroke, on the render thread.

This moves the query normalization to where the filter is built and replaces the stream with nested loops. The resource side normalization is untouched, so results are identical. `count` already does its work at parse time, so this follows the existing shape.

Measured in a dev client on a network with 37,377 resources (a Creative Storage Disk filled with the debug stick, far beyond a normal network). Cost per keystroke, including the sort. Clearing the box runs the sort alone and costs 7.8 ms, so the filter is the part above that.

| query | before | after |
|---|---|---|
| `pickaxe` | 4.5 ms | 3.8 ms |
| `@minecraft` | 15.8 ms | 10.7 ms |
| `#tool` | 21.6 ms | 14.5 ms |

Attribute and tag searches gain the most, because there the per resource stream is removed as well as the normalization being hoisted. Row counts were identical on every key press, and the existing `GridQueryParserImplTest` tests pass unchanged.

<details>
<summary>How to reproduce the measurement</summary>

Build a network with a Disk Drive, a Grid and a Creative Storage Disk, then fill the disk using the debug stick while crouching. That gives roughly 37,000 resources.

Apply this to `AbstractGridContainerMenu` for a timing per key press in the log, then type `#tool` one character at a time with and without this PR. Both the parse and the filtering are inside the timer, because this change moves work from the filtering (once per resource) to the parse (once per keystroke), and timing only the filtering would hide that.

```diff
     private boolean onSearchTextChanged(final String text) {
         try {
-            repository.setFilterAndSort(andFilter(QUERY_PARSER.parse(text), createBaseFilter()));
+            final long start = System.nanoTime();
+            repository.setFilterAndSort(andFilter(QUERY_PARSER.parse(text), createBaseFilter()));
+            LOGGER.info("query={} rows={} ms={}", text,
+                repository.getViewList().size(), (System.nanoTime() - start) / 1000000.0);
             return true;
         } catch (GridQueryParserException e) {
             repository.setFilterAndSort((v, resource) -> false);
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 -   Detector emitting an incorrect redstone signal for a brief moment on chunk load, because network storages weren't fully initialized yet. The Detector now only changes its redstone signal after the change has been stable for 1 second.
 -   Action buttons in amount screens going out of bounds in certain languages.
 -   "Craft" text in Grid going out of bounds in certain languages.
+-   Slow Grid searching on networks with many different resources.
 
 ## [3.2.1] - 2026-06-07
 
--- a/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParser.java
+++ b/refinedstorage-common/src/main/java/com/refinedmods/refinedstorage/common/grid/query/GridQueryParser.java
@@ -21,7 +21,6 @@ import com.refinedmods.refinedstorage.query.parser.node.UnaryOpNode;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -166,11 +165,17 @@ public class GridQueryParser {
         final Set<GridResourceAttributeKey> keys,
         final String query
     ) {
-        return (repository, resource) -> keys
-            .stream()
-            .map(resource::getAttribute)
-            .flatMap(Collection::stream)
-            .anyMatch(value -> normalize(value).contains(normalize(query)));
+        final String normalizedQuery = normalize(query);
+        return (repository, resource) -> {
+            for (final GridResourceAttributeKey key : keys) {
+                for (final String value : resource.getAttribute(key)) {
+                    if (normalize(value).contains(normalizedQuery)) {
+                        return true;
+                    }
+                }
+            }
+            return false;
+        };
     }
 
     private static String normalize(final String value) {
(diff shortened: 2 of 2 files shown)


raoulvdberge:
thanks!

